PR #2, opened 2021-11-02
Use a custom Less fork with slight modification

agjohnson:
This will use a slight modification to how Less `@import (optional)`
works. As described in the Less docs, optional imports silently discard
errors with missing files. Unfortunately, the implementation also
discards *all other errors*. The change in the commit below allows Less
to continue if there is a file error, but not if there is a parsing
error. This makes it far easier to discover underlying issues with
optionally included files, which is pretty much all of the SUI theme.

Commit is at:
https://github.com/agjohnson/less.js/commit/64179a24eef7308dcfb730bf0fcc3753e60ba026

This also runs npm update, which grabbed an updated Babel and friends.

You'll need to run `npm ci` locally after this is merged.

Here's what an example error will look like when it's not discarded because it was imported with `@import (optional)`:
image

So, you can now see the underlying file that caused the error and where in the file there error was, instead of weird behavior and missing variables.


astrojuanlu:
Have we tried contributing this upstream? Don't know whether Less.js is maintained, but it looks like at least it has some recent activity.

nienn:
This throws much more useful errors! I think it's safe to merge for now.

In addition, we can also try @astrojuanlu. If for some reason that does not work and it becomes a nuisance to maintain a separate branch from less.js, we can revert back to the official branch in production, as this is more useful at our current stage.

nienn (approved):

PR #3, opened 2021-11-17
Templates: Product Page

nienn:
Branched out of #1 as it uses much of the same styling and file structure developed there.


nienn:
There are tree main sections proposed in the current template at this stage
* An infographic at the top - We'll still need to draw this.
* A user segment section.
  This is the section where we can separate the users by segments and point them in different directions for the next few steps.
* A features section.
  An informative and extensive section meant for any user looking for specific or detailed information. We can use this to extend on most RTD features.


_Preview of the current template:_

product-mockup-preview

examples/mockups/product.html
```diff
@@ -0,0 +1,466 @@
...
+                whether that is Git, Mercurial, Bazaar, or Subversion, we will
+                automatically build your docs so your code and documentation are
+                never out of sync.
+              </p>
+            </div>
+            <div class="extra content">
+              <div class="ui labels">
+                <span class="ui label basic large"><i class="icon primary github"></i> GitHub</span>
+                <span class="ui label basic large"><i class="icon primary gitlab"></i> GitLab</span>
+                <span class="ui label basic large"><i class="icon primary bitbucket"></i> BitBucket</span>
```
agjohnson, line 201: The card display works great here, and the icon labels are a nice touch to avoid making this such a prose heavy page -- it provides a bit to break up the content.

It's hard to make the labels not look like buttons though. Any opinions on not using `basic label` on these?

image

That's `circular label`, `label`, and `inverted primary label`. I do like the `basic label` also though, it doesn't distract from the important parts of the content.

src/sui/themes/rtd-site/globals/site.overrides
```diff
@@ -23,8 +23,9 @@
 /*******************************
              Footer
 *******************************/
-footer {
+#footer {
```
agjohnson, line 26: Perhaps `footer, #footer` is best, both should be addressable selectors

src/sui/themes/rtd-site/globals/site.overrides
```diff
@@ -23,8 +23,9 @@
...
   color: @footerColor;
+  margin-top: 0;;
```
agjohnson, line 28:
```suggestion
  margin-top: 0;
```

examples/mockups/product.html
```diff
@@ -0,0 +1,466 @@
...
+                No additional configuration is required.
+              </p>
+            </div>
+            <div class="extra content">
+              <div class="ui labels">
+                <span class="ui label basic large"><i class="icon primary code"></i> html</span>
+                <span class="ui label basic large"><i class="icon primary file alternate outline"></i> pdf</span>
+              </div>
+            </div>
+          </div>
```
agjohnson, line 221: What do you think about adding space to these cards to show a graphic element -- ie a minimal terminal/browser graphic illustrating the workflow? Cards can have image headers, or in the body.

This seems like it would probably be a follow up project to this page and we should open an issue to track. We probably don't yet want to get bogged down on producing graphics for all of the features -- and we'd have to decide if we are producing graphics for all features/steps, or just a select couple, or what that really looks like when cards in a row have vastly differing heights due to graphics.

examples/mockups/product.html
```diff
@@ -0,0 +1,466 @@
...
+                    <div class="description">Automatic deploy</div>
+                  </div>
+                </div>
+                
+                <div class="item">
+                  <i class="icon check circle green"></i>
+                  <div class="content">
+                    <div class="description">Multiple Versions</div>
+                  </div>
+                </div>
```
agjohnson, line 165: Another piece we will likely want in this page is a the ability to call out to a feature of RTD for business. We don't have a succinct adjective for this currently, "premium" is not something we've used yet, but for example:

image

Maybe label isn't the best element there, due to the line height on the list. A link might be enough, or something like a popup? Anyways, something to maybe think about short term and long term. It's probably something we'll end up wanting to include on the community site. We'll likely also want to call out to "Gold" features -- like disabling ads.

agjohnson (commented):
This is looking great! I didn't comment on the content of the mockup yet, but the layout I think will work for the content that we want to include here.

I raised a couple of additions we might want to think about now, or plan on doing later. I'm assuming we'll probably want something like a label for plan features, but that is also something maybe more applicable to the pricing page.

Also, I couldn't comment on the removed file, there was an extra file added, `/src/sui/theme.config`, which can be removed

src/sui/themes/rtd-site/globals/site.overrides
nienn: I went for an ID as the styles applied are very specific to this `footer`.
I know it's not a common practice to have multiple footers in one page but it's perfectly acceptable to do so, should we at some point, on a different element, opt for a logical structure on which that made sense.

From [W3C](https://html.spec.whatwg.org/multipage/sections.html#the-footer-element)
> The footer element represents a footer for its nearest ancestor sectioning content or sectioning root element.

examples/mockups/about.html
```diff
@@ -0,0 +1,237 @@
+<html>
```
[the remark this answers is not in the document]
nienn: Ups, yes, I've been importing it since the first PR, with content moved from the landing page, and forgot it here.

examples/mockups/product.html
```diff
@@ -0,0 +1,466 @@
...
+
+    <!-- Contents 
+    ----------------------------------- -->
+
+    <!-- How it works -->
+    <section class="ui vertical segment very padded" id="how-it-works">
+      <div class="ui container center aligned">
+        <h1 class="ui header massive"><<i>missing</i>> Infographic</h1>
+        <p>
+          A very informative and detailed infographic that we still need to draw,<br>explaining the full Read the Docs workings, dependencies and features.
```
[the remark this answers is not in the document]
nienn: 👍 Yes! It would definitely be great to start brainstorming here and I agree with the direction you are thinking for the graphics and flow chart.

examples/mockups/product.html
```diff
@@ -0,0 +1,466 @@
...
+        </p>
+      </div>
+    </section>
+
+    <!-- User Segments -->
+    <section class="ui vertical segment inverted padded" id="user-segments">
+
+      <div class="ui container">
+        <div class="ui grid stackable three column">
```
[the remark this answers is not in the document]
nienn: > Are we trying to distill down the product page and direct different classes of users to product pages with more details?

Yes. That's my intention here. Ideally each section will clearly speak to a different user segment - _like discussed here: [user-stories.rst](https://github.com/readthedocs/readthedocs.org/blob/nienn/design-doc-landing-user-stories/docs/development/design/ext-theme/landing-page/user-stories.rst)_ - and point them in 3 different directions, to a second product page for each, where we'll explain in-depth the workings and workflow based on the group needs and tools.

examples/mockups/product.html
```diff
@@ -0,0 +1,466 @@
...
+              <h3 class="ui header"><<i>missing</i>> Illustration</h3>
+              <h3 class="ui header">Docs as Code</h3>
+              <p>
+                Read the Docs simplifies the entire process of building and deploying developer documentation. With support for Sphinx and Mkdocs, you can integrate your code and user-facing documentation using the same tools. Create beautiful documentation easily with themes, and preview every commit with Pull Request previews.
+              </p>
+              
+              <div class="ui relaxed divided list">
+
+                <div class="item">
+                  <i class="icon check circle green"></i>
```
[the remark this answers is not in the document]
nienn: Yes. I feel we are missing some alignment and refinement here for them to completely work, but that something that will make more sense to do after the work on the final content.

examples/mockups/product.html
```diff
@@ -0,0 +1,466 @@
...
+                    <div class="description">Automated Tests</div>
+                  </div>
+                </div>
+
+              </div>
+
+            </div>
+
+            <div>
+              <a href="" class="ui bui button basic inverted fluid large">About the engineering workflow</a>
```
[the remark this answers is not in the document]
nienn: > Though, what do you think about matching the button style with the icons?

It's more colorful. It can work. I changed it for now to test but I think this is another thing we'll probably want to come back and review after the content is final, in this case, especially the 3 images and infographic above as we'll probably need to balance the splash of color here with the one above.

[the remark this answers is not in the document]
nienn: I also tweaked the green a little to one a bit less electric. Probably would not go as dark as yours to keep it balanced with the brand blue. What do you think of this one? I think it works both on light and on dark backgrounds.

Screenshot 2021-12-03 091404

examples/mockups/product.html
nienn: > I do like the basic label also though, it doesn't distract from the important parts of the content.

Yes. I agree with you here. We can also explore other options but I don't mind that the labels look like buttons, as long as they are not distracting too much. I think we'll end up with more labels than the ones I used as an example and the multiple buttons per card will help removing the feeling of a call for an action. Also, maybe we can add some action: A tooltip explaining the label concept/tool?

src/sui/themes/rtd-site/globals/site.overrides
nienn: VS Code automatically adds the same coma at the end that I already added on automatic mode...  :sigh:
Guess I could change this but decided to try and get used to it.

examples/mockups/product.html
nienn: > What do you think about adding space to these cards to show a graphic element -- ie a minimal terminal/browser graphic illustrating the workflow?

I'm all in favour of adding more illustrative graphic to help us explain. We do have the obvious problem of producing these graphics... Up to a point I think it's helpful to have a more generic graphic that we can find on some image or icons pack to try illustrating ideas. If we start adding graphics to every block then at some point these generic images start becoming more noise than help and we would benefit with images produced specifically to explain our concepts.

> Cards can have image headers, or in the body.

Yes, images are one of the main elements in cards actually: https://semantic-ui.com/views/card.html

> This seems like it would probably be a follow up project

I agree. My idea for now, so we can keep moving forward defining our messaging, would be to use the more generic images where we can for now and produce just a very small number of images at this point - the infographic being one of those. Later we can come back and produce or modify images that might seem a bit out of place.

examples/mockups/product.html
nienn: I really like that basic label used as a badge for "premium"! We can explore other options but I like the effect.
Probably we'll want that badge to have the primary color we'll settle on for the corporate site. For now I'm assuming @orange.

examples/mockups/product.html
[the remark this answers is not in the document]
agjohnson: Makes sense! We can revisit in a follow up PR

examples/mockups/product.html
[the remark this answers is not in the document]
agjohnson: Yeah, once we get the features in there, the check icon might look a bit much, but there are plenty of good substitutes

examples/mockups/product.html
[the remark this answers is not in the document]
agjohnson: > this is another thing we'll probably want to come back and review after the content is final

Yeah, absolutely. I wasn't sure about having the three buttons be a pretty prominent green. It might depend on how badly we want the user to go to the child pages as well, if it's a major push then more prominent styles might help us. Agreed we can tweak after some content, I'm not particular about this change

examples/mockups/product.html
agjohnson: > we start adding graphics to every block however, at some point these generic images become noise

Yeah absolutely. I was picturing not all (probably most?) blocks would not have graphics. I definitely agreed that would be too much effort to maintain currently. I do see that only using graphics on some blocks could feel awkward however, given the two column flex layout -- unless we put two graphical features in a row, that would work well.

> Later we can come back and produce or modify images that might seem a bit out of place.

Agreed. Let's add an issue and come back to this one.

agjohnson: https://github.com/readthedocs/site-community/issues/8

examples/mockups/product.html
agjohnson: Or perhaps even calling out to the plan that the feature starts on? `Advanced plan` or `Advanced` In this case color coded to the pricing page plans even.

For ultimate fanciness, a plan color coded label that expands out to the plan name: `Advanced plan`

src/sui/themes/rtd-site/globals/site.overrides
agjohnson: Makes sense.

I do tend to use selectors that aren't just the element id. Though not particularly helpful in this example (`footer#footer`), in an example like `section#getting-started`, it can help provide some additional context to the element we're targeting without having to look it up in the html each time.

examples/mockups/product.html
agjohnson: Yeah it's definitely a toss up for me, they all look like buttons in some way. +1 on using basic button for now, and we can perhaps try again once we know what each section list would consist of.

We will be leaning into the text here, there are plenty of options for each block that won't have a great corresponding icon.

agjohnson:
I'm going to go ahead and merge this one, it sounds like we're ready to discuss content

PR #5, opened 2021-12-01
Add automatic linting check

agjohnson:
This uses prettier and all of the opinionated defaults from prettier.
I'm not always a fan of some of the decisions, but it allows us to not
think about code style.

This enables CircleCI checks on branches and pull requests. It's also
usable locally.

Also drop fomantic-ui dependency to avoid building sui and pin node for CircleCI


agjohnson:
I expect two failures here: the linting will fail because this PR doesn't actually reformat the code, that should be done on other PRs to reduce conflicts. Also, the build assets weren't updated with the pull requests yet, so that check will also fail with the assets being out of date. This PR is fine to merge even with those checks failing.

nienn (approved):

PR #10, opened 2021-12-03
Code example for developer product page

agjohnson:
Meta: I used pygmentize to generate the html for the mockup

% pygmentize -o test-rtd-yaml.html test-rtd.yaml

So we could mostly generate the html from various source files if we want to
make this really maintainable.


humitos:
> 404/5XX error pages

In particular for the 404 pages, I'd love to add more contextual data to them like a link to the project's home, at least: https://github.com/readthedocs/readthedocs.org/issues/2551

nienn:
I agree that more contextual information is useful. It's always a good idea to help the user get back to a place where he can take action, and we could do that in the same `terminal` style. I like the `terminal` analogy but it makes it even more important that it contains useful information, as I'm sure that we all see enough errors and are sometimes a bit tired of them. So the analogy needs to be kept light, useful and fun!

agjohnson:
> In particular for the 404 pages, I'd love to add more contextual data

Ah yes, that would be a great addition. I could see us having multiple 4xx/5xx error pages. The error pages related to our application and our website can be goofy/etc. But perhaps we do ultimately want a set of generic error pages for hosted documentation error pages (so customers are showing RTD branded ascii man/etc). Granted users can override these, but that would give us some ability to make our own error pages more unique

ericholscher:
Is this still a useful PR?

ericholscher:
Just pinging again here -- should we close this out, or merge it in so we have some examples to use for code blocks? We already have some config examples that are syntax highlighted, so not sure if we need this?

PR #24, opened 2022-01-12
base.html template

nienn:
The basic structure for our base template with the `topnav`, `content` and `footer` `{% blocks %}`.

I also added the option for custom `<body>` attributes and removed the inheritance from the Pelican's `!simple/base.html` and copied the file's `links` and `metadata` here for now. Later we'll probably want to add more configuration there and move the common data to an `include/`, like on [ethicalads.io](https://github.com/readthedocs/ethicalads.io/blob/main/ethicalads-theme/templates/base.html).


agjohnson (approved):
LGTM!

PR #29, opened 2022-01-18
Only remove css/js paths in theme static path

agjohnson:
This was removing images as well, which we don't want.


nienn (approved):
LGTM! We can easily keep track of old images and clean them manually so this is easier to work!

Issue #37, opened 2022-01-20
Header navigation should highlight working page


agjohnson:
It's currently static, but should show `active` class on the selected tab. Example of something similar here:

https://github.com/readthedocs/ethicalads.io/blob/main/ethicalads-theme/templates/includes/navigation.html#L26-L32

Also probably worth seeing how existing Pelican themes do this in case the above implementation was specific to EA.

nienn:
👍 this should be done as soon as I start iterating through the merged templates to add dynamic functionalities.

PR #44, opened 2022-01-26
Add href and class .active to topnav links

nienn:
Add `href` to pages that are already being developed and make the class `.active` dynamic, as expected.

_Note 1: This PR won't display anything new before the pages PRs are merged but can be merged before._
_Note 2: I also started adding the `{{ SITEURL }}` variable to internal links._

Fixes: #37

### DIFF
--- a/readthedocs-theme/templates/includes/topnav.html
+++ b/readthedocs-theme/templates/includes/topnav.html
@@ -5,10 +5,10 @@
       <div class="ui image readthedocs-logo"><img src="{{ SITEURL }}/theme/img/logo-wordmark-dark.svg"/></div>
     </a>
 
-    <a class="item{% if page and page.slug == "product" %} active{% endif %}" href="{{ SITEURL }}/product.html">Product{% if page and page.slug == "product" %} <span class="sr-only">(current)</span>{% endif %}</a>
-    <a class="item{% if page and page.slug == "pricing" %} active{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing{% if page and page.slug == "pricing" %} <span class="sr-only">(current)</span>{% endif %}</a>
-    <a class="item{% if page and page.slug == "company" %} active{% endif %}" href="{{ SITEURL }}/company.html">Company{% if page and page.slug == "company" %} <span class="sr-only">(current)</span>{% endif %}</a>
-    <a class="item{% if page and page.slug == "resources" %} active{% endif %}" href="">Resources{% if page and page.slug == "resources" %} <span class="sr-only">(current)</span>{% endif %}</a>
+    <a class="item{% if page and page.slug == 'product' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/product.html">Product</a>
+    <a class="item{% if page and page.slug == 'pricing' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing</a>
+    <a class="item{% if page and page.slug == 'company' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/company.html">Company</a>
+    <a class="item{% if page and page.slug == 'resources' %} active" aria-current="page{% endif %}" href="">Resources</a>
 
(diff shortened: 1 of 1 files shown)


readthedocs-theme/templates/includes/topnav.html
```diff
@@ -1,14 +1,14 @@
...
-      <div class="ui image readthedocs-logo"><img src="/theme/img/logo-wordmark-dark.svg"/></div>
+    <a class="header item" href="{{ SITEURL }}/">
+      <div class="ui image readthedocs-logo"><img src="{{ SITEURL }}/theme/img/logo-wordmark-dark.svg"/></div>
     </a>
 
-    <a class="item active" href="">Product</a>
-    <a class="item" href="">Pricing</a>
-    <a class="item" href="">Company</a>
-    <a class="item" href="">Resources</a>
+    <a class="item{% if page and page.slug == "product" %} active{% endif %}" href="{{ SITEURL }}/product.html">Product{% if page and page.slug == "product" %} <span class="sr-only">(current)</span>{% endif %}</a>
```
agjohnson, line 8: What's `sr-only`? Looks like a custom class to help screen readers? I would stick with ARIA attributes here instead probably.

pradyunsg: It's fairly common to add screen-reader only content with this approach -- see https://tailwindcss.com/docs/screen-readers#screen-reader-only-elements for example (Tailwind wasn't the first to do this, I think Bootstrap was).

TIL about aria-current!

agjohnson: Ahh likewise, TIL about `sr-only`. That does seem useful for some cases where supplementing graphic/visual content is maybe not a good fit for ARIA attributes alone.

I don't have enough experience using assistive technologies to say which one is better for users, but my guess would be that `aria-current` or similar would give more context to AT than just the hidden string content would. Overall, I probably lean towards ARIA as it's a supported standard, but it is handy to have `sr-only` and friends for edge cases.

nienn: I've learned a bit more about [WCAG](https://www.w3.org/TR/wai-aria-1.1/#aria-current) today too! `aria-current="page"` seems to be the better option here.

readthedocs-theme/templates/includes/topnav.html
```diff
@@ -5,10 +5,10 @@
...
-    <a class="item{% if page and page.slug == "pricing" %} active{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing{% if page and page.slug == "pricing" %} <span class="sr-only">(current)</span>{% endif %}</a>
-    <a class="item{% if page and page.slug == "company" %} active{% endif %}" href="{{ SITEURL }}/company.html">Company{% if page and page.slug == "company" %} <span class="sr-only">(current)</span>{% endif %}</a>
-    <a class="item{% if page and page.slug == "resources" %} active{% endif %}" href="">Resources{% if page and page.slug == "resources" %} <span class="sr-only">(current)</span>{% endif %}</a>
+    <a class="item{% if page and page.slug == 'product' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/product.html">Product</a>
+    <a class="item{% if page and page.slug == 'pricing' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing</a>
+    <a class="item{% if page and page.slug == 'company' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/company.html">Company</a>
+    <a class="item{% if page and page.slug == 'resources' %} active" aria-current="page{% endif %}" href="">Resources</a>
```
agjohnson, line 11: There is no href here, so we'll need to find a target here. This might even make the most sense as `Docs` and linking to our docs. I've opened #47 to address later.

agjohnson (approved):
Looks great!

Good catch on the a11y additions too! I have not been good about reviewing for these yet, so could be something we come back to and focus on later.

PR #27, opened 2022-01-13
Home page template

nienn:
Again, just moving code from mockups to templates.

* Merge after #26

PR #45, opened 2022-01-26
Example of automatic page id without a special template

agjohnson:
So far, each page we've made has a special template. Eventually, pages
will share templates though, so hardcoding this won't be as useful. This
automates `<body id="page-home">` using `page.slug` for now.

### DIFF
--- a/content/pages/homepage.md
+++ b/content/pages/homepage.md
@@ -1,5 +1,4 @@
-title: Read the Docs
-subtitle: Example subtitle
+title: Home
 template: readthedocs/homepage
 save_as: index.html
 status: hidden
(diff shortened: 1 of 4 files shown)


nienn:
The changes to the `.md` file are already being done in https://github.com/readthedocs/site-community/pull/27. Only in that case I remove the title because I prefer to see the clean site tile when at home rather than "Home - Read the Docs". This brings back the argument of using the tile slug instead of a variable separated from the title, but that's probably a separate discussion.

Other than the `.md` file, this looks good. Should we push it forward now or after the 4 pages are merged?

agjohnson:
Not needed anymore. This was mostly for CSS targeting, but we're back to native SUI for components and I haven't found the need for per-page selectors.

Issue #38, opened 2022-01-20
Add blog templates


agjohnson:
We'll need a template or two for blog content. There are maybe a few decisions lurking in this task, but let's plan to start small.

At very least, I know we'll come across whether or not we want to use image headers on every blog post. The answer here is similar to EA, and we probably do want to plan on blog posts have some graphical element at the top. Eventually, this is configurable and we have a stock of images to use at the top.

The immediate answer is: plan for an image eventually, but let's ship with just a colored background placeholder where we will use an image eventually.

EA has some helpful prior art here:
https://github.com/readthedocs/ethicalads.io/blob/main/ethicalads-theme/templates/article.html

nienn:
We also need a template `index.html` for the blog home. Not sure if you meant to include that work here. I opened a separate PR to help making issues more actionable:
* #50

PR #51, opened 2022-02-01
Blog article template

nienn:
Sets a general template to be used by the blog articles, composed of the following elements:

* Title
* Date
* Authors _(supports multiple)_
* Image _(optional)_
* Image credit _(optional)_
* Content

The image at the top is optional. It is only shown when the path it's set in the .md file. The same for the image credits.

Closes: #38


readthedocs-theme/templates/article.html
```diff
@@ -0,0 +1,49 @@
...
+  <!-- Article -->
+  <section class="ui vertical segment padded" id="article">
+    
+    <div class="ui container">
+      <div class="ui grid centered">
+        <div class="column twelve wide computer fourteen wide tablet sixteen wide mobile">
+
+
+          <h1 class="ui header">
+            <span class="ui header big">{{ article.title }}</span>
```
agjohnson, line 20: `big` is really big here, it dominates the top of the page a bit:

image

This looks a bit better, at least on the test article heading.

```suggestion
            <span class="ui header">{{ article.title }}</span>
```

agjohnson (commented):
Looks good! There are a few design issues to sort out, and it might be a good idea to test this layout with more test articles, we only have two here so far.

readthedocs-theme/templates/article.html
```diff
@@ -0,0 +1,49 @@
...
+              <img class="ui fluid image" src="{{ article.image }}">
+              {% if article.image_credit %}
+                <div class="ui basic vertical segment center aligned">
+                  <p class="ui"><i>{{ article.image_credit }}</i></p>
+                </div>
+              {% endif %}
+          {% endif %}
+
+          <div class="ui basic vertical segment padded">
+            {{ article.content }}
```
agjohnson, line 39: So, I noticed a display issue that I was expecting on the blog posts. The Pelican HTML output is not going to have the structure that would be expected -- so for example headings nested in `div.section` do not have a top margin.

image

At very least, we should catch this and apply additional top padding. It might also help to be testing against more that what we have for test blog posts. Currently, we just have two short blog posts we're testing on.

nienn: I haven't started messing with custom styles here but both the blog post and the blog index pages could definitely use some custom CSS. I opened an issue for the blog post here #67; had already one open for the blog index #61

nienn: * PR #70 fixes this and closes #67

agjohnson: There might be more work to do later, but this PR should probably address some of the basic display issues on blog content. Not having more test blog posts here is probably delaying finding these issues

readthedocs-theme/templates/article.html
```diff
@@ -0,0 +1,47 @@
...
+            <div class="ui basic vertical segment left floated">{% for author in article.authors %}{{ author }}{% if not loop.last %}, {% endif %}{% endfor %}</div>
+            <div class="ui basic vertical segment right floated"><time datetime="{{ article.date.isoformat() }}">{{ article.locale_date }}</time></div>
+          </div>
+
+          <img class="ui fluid image" src="{{ article.image or '/theme/img/defaults/alesia-kazantceva-VWcPlbHglYc-unsplash.jpg' }}">
+          {% if article.image and article.image_credit %}
+            <div class="ui basic vertical segment center aligned">
+              <p class="ui"><i>{{ article.image_credit}}</i></p>
+            </div>
+          {% endif %}
```
nienn, line 34: If we decide to show attribution here this might change to something like this

```suggestion
          {% if (article.image and article.image_credit) or not article.image %}
            <div class="ui basic vertical segment center aligned">
              <p class="ui"><i>{{% if not article.image %}Photo by Mailchimp{% else %}{{ article.image_credit}}{% endif %}</i></p>
            </div>
          {% endif %}
```

nienn:
https://github.com/readthedocs/site-community/pull/51/commits/474782e790d97b682810996e976b56fa541c0066 adds a default image to the article that is shown in case no image is supplied by the file meta.
I'm tentatively adding default template images to folder `/readthedocs-theme/static/img/defaults/` which bundles into `/theme/img/defaults/`.

The image I'm using: https://unsplash.com/photos/VWcPlbHglYc. Not sure if we want to add attribution --added related [comment above](https://github.com/readthedocs/site-community/pull/51/files#r806191913).

nienn:
https://github.com/readthedocs/site-community/pull/51/commits/f2c394943cfa0095eabc2fbda6d875a1987f4d78 styles images inside blog posts.

Expects images in the following formats:

1. .md:
    ` An image `
    
    _outputs_
    ```html
    <p>
       <img alt="An image" src="/images/posts/image.jpg">
    </p>
    ```

2. .srt:
    ` .. image:: {static}../images/posts/image.jpg `
    
    _outputs_
    ```html
      <img alt="An image" src="/images/posts/image.jpg">
    ```


3. ```html
    <figure>
      An image alt.
      <figcaption>An image caption.</figcaption>
    </figure>
    ```

4. ```html
    <img src="{static}../images/posts/image.jpg" alt="An image alt.">
    ```

readthedocs-theme/templates/article.html
```diff
@@ -0,0 +1,47 @@
...
+            <span class="ui header">{{ article.title }}</span>
+          </h1>
+
+          <div class="ui basic vertical segment clearing">
+            <div class="ui fitted divider"></div>
+            <div class="ui basic vertical segment left floated">{% for author in article.authors %}{{ author }}{% if not loop.last %}, {% endif %}{% endfor %}</div>
+            <div class="ui basic vertical segment right floated"><time datetime="{{ article.date.isoformat() }}">{{ article.locale_date }}</time></div>
+          </div>
+
+          <img class="ui fluid image" src="{{ article.image or '/theme/img/defaults/alesia-kazantceva-VWcPlbHglYc-unsplash.jpg' }}">
```
agjohnson, line 29: Probably makes more sense renamed to a generic image name, so that we can easily replace it in the future.

Also, `readthedocsext/theme` is the output path of webpack, it might make sense to keep all of the blog post header images together in a single path

```suggestion
          <img class="ui fluid image" src="{{ article.image or '/images/posts/default.png' }}">
```

readthedocs-theme/templates/article.html
agjohnson: Yeah, we should show attribution with a link to the artist.

EthicalAds shows full attribution and a link:
image

The template logic in your example might be a little off though. It seems it would show an empty `<p>` if there is an `article.image` but no `article.image_credit`.

agjohnson (commented):
This looks good locally so far. We do want attribution on the image, so should include that here as well.

readthedocs-theme/templates/article.html
nienn: I agree with the image name. I used the original so we could keep track of the source, but since we are using a full attribution that's not needed. I'll also move it to the blog `images/posts/` folder.

readthedocs-theme/templates/article.html
nienn: Adding the full attribution. Although this might change if we go for the generative geometric image.

```suggestion
          {% if (article.image and article.image_credit) or not article.image %}
            <div class="ui basic vertical segment center aligned">
              <p class="ui">
                <i>
                  {% if not article.image %}
                    Photo by <a href="https://unsplash.com/photos/VWcPlbHglYc">Mailchimp</a> on <a href="https://unsplash.com/">Unsplash</a>
                  {% else %}
                    {{article.image_credit}}
                  {% endif %}
                </i>
              </p>
            </div>
          {% endif %}
```

For reference, this is how this looks:

printscreen-16-2-2022

> The template logic in your example might be a little off though. It seems it would show an empty `<p>` if there is an `article.image` but no `article.image_credit`.

```
{% if (article.image and article.image_credit) or not article.image %}
```
I'm checking if _both_ the article.image `and` article.image_credit are set, in which case I use the article.image_credit bellow, `or` if the article.image does `not` exist, in which case I show the default image credit bellow.

readthedocs-theme/templates/article.html
nienn: Looking further at the [Pelican structure](https://docs.getpelican.com/en/latest/themes.html#structure) for themes I notice that images inside the folder `/content/images/posts/` are only built if they are referenced from a content file. Images that are used _(only)_ by the templates must be placed inside the template folder structure, which is handled by Webpack.

agjohnson: I didn't have this issue. You might find the images are not copied immediately when using `npm run dev` though. Saving a template file does copy files from `/images/posts/default.png` to the output path. So, I think `/images/posts/` is still the correct path.

readthedocs-theme/templates/article.html
agjohnson: Sorry, I was talking about the second conditional:

```{{% if not article.image %}Photo by Mailchimp{% else %}{{ article.image_credit}}{% endif %}```

You're right though that the parent conditional protects this nested conditional. It's just a little confusing due to the similar conditional. I get the logic you're going for now though.

readthedocs-theme/templates/article.html
nienn: https://github.com/readthedocs/site-community/pull/51/commits/f22cd45de8a94c54a127c141c2fdda4c8621637e moves the default image to the `/content/images/posts/` folder.

I'm still not getting the correct build thought.
Running `npm run build` then `npm run dev`: the `test.jpg` image, that is referenced from the `.md` and `.rst` files, is built into the `/output` folder but the `default.jpg` is not. Tried several other things with always the same result.

agjohnson: Seems you didn't rebase on d7990011d065e321d9fffd3508a75bc865876d5a after merging, so none of the images in this path would get picked up at all.

agjohnson: I pushed up eeea54d to resolve this.

agjohnson: Loads correctly now: 
image

PR #57, opened 2022-02-02
Test import .rst article from old blog

nienn:
Importing this file from the current blog into the new structure, found two issues:

* The file `metadata` had to be edited to fit [Pelican](https://docs.getpelican.com/en/3.6.3/content.html#file-metadata) purposes.
* `Cross-References` are also producing errors. Removed those and build passed with no more problems.


agjohnson:
We could use more blog posts port over to help mature the blog templates a bit more. Blog posts that are heavy on internal linking or intersphinx linking will require the most work to port over. It would be most helpful to keep the same file names

ericholscher:
Gonna close this out, since we aren't prioritizing this work.

PR #58, opened 2022-02-02
Blog home template

nienn:
#### Notes

* I'm chaining this PR on #51 because it uses changes made there to the `test.md` file.

* The current template hasn't got any specific styling, it's using just the native features from Semantic UI. In future, the UX could be improved by adding custom CSS and improving the default look and interaction.

#### Todo

This template requires `pagination.html`, which doesn't yet exist.

### DIFF
--- a/readthedocs-theme/templates/index.html
+++ b/readthedocs-theme/templates/index.html
@@ -1,0 +1,111 @@
+{% extends "base.html" %}
+{# This is the blog index, not the home page template #}
+
+
+{% block content %}
+
+  <!-- Intro -->
+  <section class="ui vertical segment" id="blog-intro">
+    <div class="ui container">
+      <div class="ui segment basic">
+
+        {% block content_title %}
+          <div class="ui basic vertical big segment">
+          <h1 class="ui header">
+            <span class="ui header">Read the Docs Blog</span>
+          </h1>
+            <p>
+              Short intro
+            </p>
+          </div>
+        {% endblock %}
+
+      </div>
+    </div>
+  </section>
+
+
+  <section class="ui vertical segment" id="articles-list">
+    <div class="ui container">
+      <div class="ui segment basic">
+
+        <!-- Latest  2 articles -->
+        <div class="ui two cards stackable">
+          {% for article in articles_page.object_list %}
+            {% if not articles_previous_page and loop.index <= 2 %}
+              <a class="ui card" href="/{{ article.url }}">
+                <div class="image">
+                  <img src="{{ article.image }}">
+                </div>
+                <div class="content">
+                  <div class="header">{{ article.title }}</div>
+                  <div class="meta">
+                    <span class="date"><time datetime="{{ article.date.isoformat() }}">{{ article.locale_date }}</time></span>
+                  </div>
+                  <div class="description">
+                    {% if article.description %}
+                      {{ article.description|striptags|escape }}
+                    {% else %}
+                      {{ article.summary|striptags|escape }}
+                    {% endif %}
+                  </div>
+                </div>
+                <div class="extra content">
+                  <div>
+                    {% for author in article.authors %}{{ author }}{% if not loop.last %}, {% endif %}{% endfor %}
+                  </div>
+                </div>
+              </a>
+            {% endif %}
+          {% endfor %}
+        </div>
+
+        <!-- Articles list -->
+        <div class="ui three cards stackable">
+          {% for article in articles_page.object_list %}
+            {% if articles_previous_page or not loop.index <= 2 %}
+              <a class="ui card" href="/{{ article.url }}">
+                <div class="image">
+                  <img src="{{ article.image }}">
+                </div>
+                <div class="content">
+                  <div class="header">{{ article.title }}</div>
+                  <div class="meta">
+                    <span class="date"><time datetime="{{ article.date.isoformat() }}">{{ article.locale_date }}</time></span>
+                  </div>
+                  <div class="description">
+                    {% if article.description %}
+                      {{ article.description|striptags|escape }}
+                    {% else %}
+                      {{ article.summary|striptags|escape }}
+                    {% endif %}
+                  </div>
+                </div>
+                <div class="extra content">
+                  <div>
+                    {% for author in article.authors %}{{ author }}{% if not loop.last %}, {% endif %}{% endfor %}
+                  </div>
+                </div>
+              </a>
+            {% endif %}
+          {% endfor %}
+        </div>
+
+      </div>
+    </div>
+  </section>
+  
+  <section class="ui vertical segment padded" id="articles-list">
+    <div class="ui container">
+      <div class="ui segment basic">
+
+        {% if articles_page.has_other_pages() %}
+          {% include 'pagination.html' %}
+        {% endif %}
+
+      </div>
+    </div>
+  </section>
+
+
+{% endblock content %}
--- a/content/posts/test.md
+++ b/content/posts/test.md
@@ -1,7 +1,11 @@
-Title: Testing Pelican
-Date: Jan 2, 2021
-description: Testing Pelican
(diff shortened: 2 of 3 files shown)


agjohnson:
I've updated the base to #51

readthedocs-theme/templates/index.html
```diff
@@ -0,0 +1,113 @@
...
+    <div class="ui container">
+      <div class="ui segment basic">
+
+        {% block content_title %}
+          <div class="ui basic vertical big segment">
+          <h1 class="ui header">
+            <span class="ui header">Read the Docs Blog</span>
+          </h1>
+            <p>
+              Short intro
```
agjohnson, line 18: I would just use one of the one-sentence intros we've used on other pages.

readthedocs-theme/templates/index.html
```diff
@@ -0,0 +1,113 @@
...
+
+        <!-- Latest  2 articles -->
+        <div class="ui two cards stackable">
+          {% for article in articles_page.object_list %}
+            {% if not articles_previous_page and loop.index <= 2 %}
+              <a class="ui card" href="/{{ article.url }}">
+                {% if article.image %}
+                  <div class="image">
+                    <img src="{{ article.image }}">
+                  </div>
```
agjohnson, line 40: This is another area where we should have blog posts with test images.

It is probably going to make me want images on all blog posts though, but i'm not sure how we make both types of posts work together on this page. EthicalAds forces images on all blog posts, which is definitely a direction we could go with the addition of a placeholder of sorts.

readthedocs-theme/templates/index.html
```diff
@@ -0,0 +1,113 @@
...
+
+  <section class="ui vertical segment" id="articles-list">
+    <div class="ui container">
+      <div class="ui segment basic">
+
+        <!-- Latest  2 articles -->
+        <div class="ui two cards stackable">
+          {% for article in articles_page.object_list %}
+            {% if not articles_previous_page and loop.index <= 2 %}
+              <a class="ui card" href="/{{ article.url }}">
```
agjohnson, line 36: The text is all underlined for me, which is not the normal styles for cards:

image

I believe this needs an extra nested `div`, see: https://semantic-ui.com/views/card.html#link-card

readthedocs-theme/templates/index.html
```diff
@@ -0,0 +1,113 @@
...
+              </a>
+            {% endif %}
+          {% endfor %}
+        </div>
+
+        <!-- Articles list -->
+        <div class="ui three cards stackable">
+          {% for article in articles_page.object_list %}
+            {% if articles_previous_page or not loop.index <= 2 %}
+              <a class="ui card" href="/{{ article.url }}">
```
agjohnson, line 69: Same note as above on the extra div.

agjohnson (commented):
Looks good, I think the card layout works well here. I'd be most curious to see this with images as well. I feel we might want to enforce images if we start putting images on any of the posts.

readthedocs-theme/templates/index.html
nienn: I'm almost always in favor of adding more graphics; although instead of enforcing, I would probably add a default image instead, to be used when no other is provided.

readthedocs-theme/templates/index.html
nienn: * Opened a PR to fix this in #69

nienn:
The last two commits add code to get the default images for the article cards list (the actual image file is added on PR #51) and style those images to be all of the same height.

Note: The properties used - ie: [`object-fit`](https://caniuse.com/?search=object-fit) - assume that the community website is not supporting `ie11`. The site will still be completely usable by `ie11` but the card images will have different heights. The cards total height is still be the same in `ie11` and they align perfectly in grid layout.

readthedocs-theme/templates/index.html
agjohnson: This still needs changed. If you're not sure on what the copy is we're putting here it can be dropped.

agjohnson (commented):
This is looking good. Pagination could use some tweaks, but the layout of the page works really well.

I did feel the default image could be tweaked once I saw it repeated in grid view. I suggested a default pattern or texture image, but perhaps even rotating through a couple images helps. I hacked up something quick here:

https://github.com/readthedocs/site-community/pull/75/files

I'd vote just use an abstract or generative image to fill in instead of a stock photo. This will only be used if someone digs down through our blog post history on the pagination, though I could also see us using a next/last post feature on each blog post, or a suggested blog post feature like EA.

pelicanconf.py
```diff
@@ -68,5 +68,5 @@
 SOCIAL = ()
 
 # Posts settings
-DEFAULT_PAGINATION = 10
+DEFAULT_PAGINATION = 9
```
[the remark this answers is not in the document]
nienn: Committing suggestion from conversation above.
```suggestion
DEFAULT_PAGINATION = 8
```

readthedocs-theme/templates/index.html
```diff
@@ -0,0 +1,107 @@
...
+    <div class="ui container">
+      <div class="ui segment basic">
+
+        <!-- Latest  2 articles -->
+        <div class="ui two cards stackable">
+          {% for article in articles_page.object_list %}
+            {% if not articles_previous_page and loop.index <= 2 %}
+              <a class="ui card" href="/{{ article.url }}">
+                <div class="image">
+                  <img src="{{ article.image  or '/theme/img/defaults/alesia-kazantceva-VWcPlbHglYc-unsplash.jpg'}}">
```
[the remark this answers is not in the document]
nienn: This is dependant on the [resolution from the other PR](https://github.com/readthedocs/site-community/pull/51#discussion_r807732883) and the actual image is included by that PR.
Presently, with commit https://github.com/readthedocs/site-community/pull/51/commits/b85bae37d352a528ba513bff711c2317945ef41e I changed the image name to `article.jpg`.

```suggestion
                  <img src="{{ article.image  or '/theme/img/defaults/article.jpg'}}">
```

readthedocs-theme/templates/index.html
nienn: From the `company.html` copy:
```suggestion
              This is where we post all the week to week progress updates and announcements.
```

readthedocs-theme/templates/index.html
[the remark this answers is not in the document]
agjohnson: Noted on the other PR, I did not have issues with using `/content/images/default.png` like you are describing. I did notice in #75 that the defaults aren't picked up immediately by inotify/Pelican when changed, but do make it into the output either by saving a different template file, restarting `npm run dev`, or I assume doing `npm run build`.

agjohnson:
We can return to the default image when importing blog posts, but I'd probably vote a texture/pattern/geometric image.

This one is chained and conflicting, so will need a merge in -- along with the filename fix above.

readthedocs-theme/templates/index.html
agjohnson: Seems fine for now!

readthedocs-theme/templates/index.html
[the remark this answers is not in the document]
nienn: https://github.com/readthedocs/site-community/pull/58/commits/6e0bb80a7e8e5176b211ee1b57861e9b49b6c636 updates the default image src to `/content/images/posts/`.
As noted in this https://github.com/readthedocs/site-community/pull/51#discussion_r808886650, I'm still not getting the correct build thought.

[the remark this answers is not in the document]
agjohnson: Probably this is the result of the PR chaining. There is no `/content/images/posts/default.jpg` on this branch:

```
%  ls content/images/posts/default.jpg
ls: cannot access 'content/images/posts/default.jpg': No such file or directory
```

PR #60, opened 2022-02-02
Update .topnav links

### LINKED ISSUE #47: Navigation: add link for Resources
The link in the top navigation current has a menu option for `Resources`, as of #44 it wasn't yet linked however.

This might also make sense as a `Docs` link that points out to our doc pages.

nienn:
Adds the missing links: **resources**, **support** and **application**.

The first two point outside the site, to the blog.
I used `target="_blank"` on these two so the user wouldn't loose the `.topnav` to navigate within the site.

Closes #47

### DIFF
--- a/readthedocs-theme/templates/includes/topnav.html
+++ b/readthedocs-theme/templates/includes/topnav.html
@@ -8,12 +8,12 @@
     <a class="item{% if page and page.slug == 'product' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/product.html">Product</a>
     <a class="item{% if page and page.slug == 'pricing' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing</a>
     <a class="item{% if page and page.slug == 'company' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/company.html">Company</a>
-    <a class="item{% if page and page.slug == 'resources' %} active" aria-current="page{% endif %}" href="">Resources</a>
+    <a class="item" target="_blank" href="https://docs.readthedocs.io/en/stable/guides/">Resources</a>
 
     <div class="right menu">
-      <a class="item" href="">Support</a>
+      <a class="item" target="_blank" href="https://docs.readthedocs.io/en/stable/support.html">Support</a>
       <div class="item">
-        <a class="ui button primary">
+        <a class="ui button primary" href="https://readthedocs.org/dashboard/">
           Open Application
         </a>
       </div>


readthedocs-theme/templates/includes/topnav.html
```diff
@@ -8,12 +8,12 @@
     <a class="item{% if page and page.slug == 'product' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/product.html">Product</a>
     <a class="item{% if page and page.slug == 'pricing' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing</a>
     <a class="item{% if page and page.slug == 'company' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/company.html">Company</a>
-    <a class="item{% if page and page.slug == 'resources' %} active" aria-current="page{% endif %}" href="">Resources</a>
+    <a class="item" target="_blank" href="https://docs.readthedocs.io/en/stable/guides/">Resources</a>
 
     <div class="right menu">
-      <a class="item" href="">Support</a>
+      <a class="item" target="_blank" href="https://docs.readthedocs.io/en/stable/support.html">Support</a>
```
agjohnson, line 14: We can also link to `https://readthedocs.org/support` too. Neither are superior experiences though. Eventually, in our application and/or documentation, we should have a bit more helpful of a portal for users. The application support page is the one we currently link to, so might be a better option for now.

I'm not sure exactly where we maintain a page like this though. It's a bit too much to maintain in our application, but we also require authentication on our application support page. For now, I think linking into the application is enough.

nienn:
```suggestion
      <a class="item" target="_blank" href="https://readthedocs.org/support">Support</a>
```

agjohnson: If we are going to open that application without blank, we could also open the support link without `_blank`

nienn: My thought on this was that the application will maintain the top navigation that will allow users to come back here while also maintain the feeling of not actually leaving the site. But I guess that's only true when we actually have the same top navigation working on the application... I'm removing the target `_blank` on both for now and setting a reminder to review this again after the app is done.

readthedocs-theme/templates/includes/topnav.html
```diff
@@ -8,12 +8,12 @@
     <a class="item{% if page and page.slug == 'product' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/product.html">Product</a>
     <a class="item{% if page and page.slug == 'pricing' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing</a>
     <a class="item{% if page and page.slug == 'company' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/company.html">Company</a>
-    <a class="item{% if page and page.slug == 'resources' %} active" aria-current="page{% endif %}" href="">Resources</a>
+    <a class="item" target="_blank" href="https://docs.readthedocs.io/en/stable/guides/">Resources</a>
 
     <div class="right menu">
-      <a class="item" href="">Support</a>
+      <a class="item" target="_blank" href="https://readthedocs.org/support">Support</a>
```
nienn, line 14:
```suggestion
      <a class="item" href="https://readthedocs.org/support">Support</a>
```

readthedocs-theme/templates/includes/topnav.html
agjohnson: Sounds good. I'd probably still rather skip _blank, as it's an internal redirect and if the user is going to our application/support, they most likely did not have any intention on browsing our site anyways.

PR #64, opened 2022-02-07
Pagination template

nienn:
Template for the pagination `{% included %}` in the blog home page. Displayed automatically when there are more pages to show. The PR also changes the maximum number of posts per page and adjusts the `index.html` structure, where it's included.

_Preview:_
Screenshot 2022-02-07 153835

* Requires #58


readthedocs-theme/templates/includes/pagination.html
```diff
@@ -0,0 +1,20 @@
+{% if DEFAULT_PAGINATION %}
+  {% set first_page = articles_paginator.page(1) %}
+  {% set last_page = articles_paginator.page(articles_paginator.num_pages) %}
```
agjohnson, line 3: It looks like we set up to do a numeric pagination, but then don't use the variables. Is numeric pagination something we want here?

https://semantic-ui.com/collections/menu.html#pagination

agjohnson (commented):
Look

nienn:
Good point. Updated the PR to remove those variables https://github.com/readthedocs/site-community/pull/64/commits/82c08d2a5812e2409503d30bbf018347e84ba5ba.

> I don't think users will find our articles by scrolling through our history like that.

I agree, we don't need numeric pagination and It would add more visual complexity to the page.

readthedocs-theme/templates/includes/pagination.html
nienn: Removed line 2 and 3 in commit https://github.com/readthedocs/site-community/pull/64/commits/82c08d2a5812e2409503d30bbf018347e84ba5ba.

agjohnson (approved):

PR #68, opened 2022-02-08
Footer: Improve links :hover color

### LINKED ISSUE #52: Footer: Improve links :hover styling
Footer links turn dark when hovered, which makes them almost the same color as the background.

nienn:
Makes footer links `@white` when hovered.

* Closes #52

### DIFF
--- a/src/sui/themes/rtd-site/globals/site.overrides
+++ b/src/sui/themes/rtd-site/globals/site.overrides
@@ -74,6 +74,7 @@ body {
   }
   a,
   a:hover {
+    color: @footerHoverColor;
     text-decoration: none;
   }
   .ui.list a.item:hover {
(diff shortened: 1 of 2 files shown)


src/sui/themes/rtd-site/globals/site.overrides
```diff
@@ -74,6 +74,7 @@ body {
   }
   a,
   a:hover {
+    color: @footerHoverColor;
```
agjohnson, line 77: This is also on the `a` selector, so will alter the default link color. Hover would not be noticeable in this case, so should probably have a dedicated selector.

nienn: 👍  Ups, good catch! That was not the intention. https://github.com/readthedocs/site-community/pull/68/commits/dd7d9232afe4a05a213e15aaba0438ca04b4de3a should fix.

agjohnson (approved):

PR #74, opened 2022-02-15
Add images path to pelicanconf.py

nienn:
Images were not being built into `/output/images/` as expected.
This adds the images path to pelicanconf.py `STATIC_PATHS`


agjohnson (approved):
Glad that was a simple fix!

PR #75, opened 2022-02-16
Generative default image example

agjohnson:
I was playing around with a script that I had for generating some images, though plenty of other options on unsplash too. Just squirreling this away for now, but perhaps this is some direction. Multiple images looks a little better if the listing page is going to be all image-less or all old posts.

I'm not yet sure on the size of the lead image though, the wide container does make a `ui fluid image` take up most of the viewport on a 4:3 aspect image. I sort of like the stout image on the article page, and the article list page still looks fine.

image

image


readthedocs-theme/templates/article.html
```diff
@@ -26,13 +25,11 @@ <h1 class="ui header">
...
           </div>
 
-          {% if article.image %}
-              <img class="ui fluid image" src="{{ article.image }}">
-              {% if article.image_credit %}
-                <div class="ui basic vertical segment center aligned">
-                  <p class="ui"><i>{{ article.image_credit }}</i></p>
-                </div>
-              {% endif %}
+          <img class="ui fluid image" src="{{ article.image  or '/images/posts/default-{:.0f}.png'.format(article.date.strftime('%Y%m%d') | int % 7) }}" />
```
agjohnson, line 28: Just a hack to load one of 7 images via modulus operation.

nienn:
I like the idea and the look of these images!
I don't have strong arguments against or in favor of using them over traditional figurative images, just the following considerations:

**Pros for generative geometric images:**
* Fun and attractive look
* Techie look, which relates with our audience

**Pros for figurative images:**
* Create an emotional bond between article and user. The user comes to associate the image with what he read on the article. This often helps categorizing concepts during the process of reading and interiorization.

agjohnson:
Yeah, for *new* blog posts I think using unique images from unsplash, or at least using unique images for most new posts, makes more sense. This is mostly for historical posts, which won't have images applied because that wasn't part of the existing blog content.

PR #78, opened 2022-02-16
Open source is not a proper noun, it can be lower case

agjohnson:
I'm not convinced that should be there either way though, RTD is a
company, not just an open source project.


nienn (approved):
Looks good!

Issue #76, opened 2022-02-16
Finish product page card button lists


agjohnson:
The lists have placeholder icons right now:

![image](https://user-images.githubusercontent.com/1140183/154214575-e94f531b-23e1-4f7b-b883-1d9f755b373c.png)

PR #81, opened 2022-02-16
Add card label icons in the product page

nienn:
Adds the missing icons (previously placeholders) to the labels on the product page cards.

screencapture-localhost-8080-product-html-2022-02-16-18_43_10

_Note: sadly, sui has no icon remotely similar to a branch/git/version-control. I don't love the current option but I think it's the most closely related_

Closes: #76

### DIFF
--- a/readthedocs-theme/templates/readthedocs/product.html
+++ b/readthedocs-theme/templates/readthedocs/product.html
@@ -173,7 +173,7 @@
           </div>
           <div class="extra content">
             <div class="ui labels">
-              <span class="ui label basic large"><i class="icon primary hourglass outline"></i> Version control</span>
+              <span class="ui label basic large"><i class="icon primary copy outline"></i> Version control</span>
             </div>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


readthedocs-theme/templates/readthedocs/product.html
```diff
@@ -173,7 +173,7 @@ <h4>Multiple versions</h4>
           </div>
           <div class="extra content">
             <div class="ui labels">
-              <span class="ui label basic large"><i class="icon primary hourglass outline"></i> Version control</span>
+              <span class="ui label basic large"><i class="icon primary copy outline"></i> Version control</span>
```
agjohnson, line 176: Perhaps:

```suggestion
              <span class="ui label basic large"><i class="icon primary code branch"></i> Version control</span>
```

agjohnson: No, the icon I'm describing should be:
image

I get the correct icon that I'm describing:
image

nienn: Aha, that one is perfect! It's exactly the one I was looking for. I was looking on the Semantic UI docs thought. Glad to see that Fomantic updated it. Makes a lot more sense than the SUI one.

agjohnson: Odd, not sure where you were seeing that, I see the same on on SUI's docs:

image

nienn: I was using these docs: https://semantic-ui.com/elements/icon.html#code.
Now changed to the Fomantic equivalent after realising this.

agjohnson: ah yeah, so that grid view is wrong. the search shows the correct icon.

PR #92, opened 2022-02-24
Pricing update

humitos:
Gold is not related to Advanced plan.

### DIFF
--- a/readthedocs-theme/templates/readthedocs/pricing.html
+++ b/readthedocs-theme/templates/readthedocs/pricing.html
@@ -16,14 +16,14 @@
           <span class="ui header large">for Business</span>
         </h2>
       </div>
-      
+
       <div class="ui three stackable cards">
 
         <!-- Card -->
         <div class="ui inverted card">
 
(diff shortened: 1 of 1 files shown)


nienn:
The item is still needed but it's pointing to the wrong plan, probably left behind by a copy/past.
It's meant to read _**"All the features from the `Base` plan"**_ --instead of `Advanced` plan.

humitos:
That makes sense. In that case, would we need the same sentence for Ad-free as well?

nienn:
The Ad-free plan already states that in its text:

Screenshot 2022-02-28 084116

It's in sentence style, instead of the list style the other plans use, to reinforce the idea of being a special case and not a plan in itself.

agjohnson (approved):

agjohnson:
I've updated with the conversation here.

- 0642b958692a6418e92e9d8fc816698d55d09b04 fixes copy some
- 18e60c1 cleans whitespace

nienn (approved):
LGTM!

PR #96, opened 2022-03-08
Add favicon images and links

### LINKED ISSUE #87: Site is missing favicon
The hosted version has no favicon. We have several examples of this already which could be imported. The last time we redid the favicon was probably 5+ years ago though, so perhaps there are some better standards to support.

nienn:
Screenshot 2022-03-08 183555

Closes #87


agjohnson:
The 16x16 looks about what we expected -- very scaled down and the scaling makes it look clipped at the top. The current readthedocs.com favicon looks better, if we fixed the halo.

agjohnson:
I pushed a fix at #98

agjohnson:
What's the best standard to use for `favicon.ico`? Should this be the 16x16 format?

agjohnson:
Seems the answer is a multiformat ico file, 16x16, 32x32, and 48x48 formats. We can come back to this, as I'm not really even sure what would use this format over the explicit definitions for favicon sizes.

agjohnson (approved):
Pending some testing with favicon.ico, but we can come back to that

PR #98, opened 2022-03-10
Fix 16x16 favicon

agjohnson:


nienn:
Should we account for browsers in dark mode? I made the other icon versions a few shades lighter so it's still be visible.
 
Current:
darker

Hex #4d515b:
lighter

agjohnson:
Good point, I'll update!

Is your browser tab icon 16x16, or is it something like 24x24? The 16x16 icon looks like it's scaled out a little bit more than i'd expect.

agjohnson:
No, it's not GitHub, your screen grab isn't scaled down for me. It does look like your browser is using a 24x24 -- I copied the 16x16 in for reference.

image

24x24 is probably another format to include, probably from my source.

nienn:
My browser uses the 16px favicon. The mockup above was not a print-screen however, I grabbed the image here and mounted it in Photoshop. Not sure why the discrepancy.

agjohnson:
Yeah, I think it's using the 16x16 icon scaled up to 24x24 though, as there isn't a closer icon size. I bet if you add a 24x24 size favicon definition your browser icon will use that, and it might look a bit nicer.

nienn:
The 24x24 icon might be useful for the pinned site browser UI. Chrome uses icons multiple of 48px. The 32x32px favicon is currently supported by most browsers and the recommended size thought Chrome scales it down to 16px.

https://stackoverflow.com/questions/4014823/does-a-favicon-have-to-be-32%C3%9732-or-16%C3%9716

nienn (approved):

agjohnson:
Yeah, I'm just guessing that the 24x24 will look better, I haven't tested this. Your browser is definitely using the 16x16 scaled up to something larger though, you can tell because the shape and positioning is different in the 16x16/24x24 version than it is in the 32x32 and beyond.

Issue #105, opened 2022-03-10
Front page how it works section placement and spacing below 990px


agjohnson:
![image](https://user-images.githubusercontent.com/1140183/157766552-d02e35fc-7de5-451b-bced-9472014f0b9e.png)

The easiest option might be to limit the width of the text containers at this viewport size. There will still be a lot of empty space to the sides of the graphics/text, but it will at least be aligned

nienn:
Same as on https://github.com/readthedocs/site-community/issues/103#issuecomment-1066600392. There's no native SUI option for selectively centering text on different viewport so we might need some custom CSS. Also, this brings up a new question: should we target each case individually or create viewport sensitive text centering classes to reuse on different situations?

Narrowing the columns max width is easy enough.

Issue #104, opened 2022-03-10
Footer: menu order and placement below 990px breakpoint


agjohnson:
The footer menu shifts slightly out of order below 990px:

![image](https://user-images.githubusercontent.com/1140183/157765530-7463d5ce-2155-4ed7-b989-6bc005dc9a31.png)

I'm not sure if that placement is correct, but it feels like it should fit on the first row at this breakpoint.

nienn:
SUI uses a hardcoded `width: 50% !important` for the `.doubling.grid` columns bellow viewport `992px`. This forces the 50% width without any of the usual flexibility from flexbox. So the placement is in accordance with the expected behaviour from SUI.

IMO, however, the footer is a custom element with very specific requirements dictated by each site. It should not be dependant entirely on the framework native behaviours.
I believe the footer should be styled with custom CSS and that would both simplify future editing at the same time as it gives finer control over behaviours.

agjohnson:
Without doubling, it looks like this:

![image](https://user-images.githubusercontent.com/1140183/158247656-6554a83f-18af-4729-aefd-d2efd7ee5bcb.png)

That seems better than with doubling, even if the text field is squashed.

nienn:
So my feeling here is that the columns are very unbalanced --the white space left by the narrower columns leaves them hanging in heavy contrast with the wider first column.

Using 3 columns with fixed widths works technically, it's just not visually appealing and leaves a messy feeling.

agjohnson:
The desktop view size is already using 3 columns though, so nothing really changes by dropping doubling. It does prevent the 3rd column from stacking under the 1st, which seems definitively worse.

I think the answer is removing doubling and tuning the content in the columns to help match the vertical footprint of the first column.

Issue #103, opened 2022-03-10
Front page image placement and header alignment


agjohnson:
Below 990px breakpoint, content reflows to a vertical display:

![image](https://user-images.githubusercontent.com/1140183/157762600-a459f217-e258-4c4d-b0e9-04811ace61fc.png)

At tablet/mobile, similar issues:

![image](https://user-images.githubusercontent.com/1140183/157763648-e34ae47d-810e-4fd1-a454-967aa1d36c42.png)

## Other options

- I think the reason it looks out of place is because it's an abstract flourish type of graphic. If it communicated more to the user, it would be fine. So a more applicable graphic is one option.
- Alternatively, just use a full bleed graphic behind the text/buttons
- We could hide the image/column on tablet/mobile viewports at least. This is the easiest option.

----

- [ ] Resolve issues with image placement
- [ ] And/or resolve issues with image scaling. Don't fluid scale out to the full display width on some tablet/mobile width.
- [ ] Resolve issues with text alignment

nienn:
I'm +1 on hiding the image on mobile. There's no real reason to make the user scroll an extra screen. I also think that all of the elements on the above screen (posted print-screens) should become center-aligned on mobile.

SUI does not come with many options for responsive behaviour thought, so we might need some custom CSS.
For hiding the image we can use the [device only](https://fomantic-ui.com/collections/grid.html#device-visibility) grid option, which requires adding complexity to the DOM structure but it works natively. There's no such options for centering the text on selected viewports.

agjohnson:
Yeah, a reusable class would probably be best. There are several places on the marketing pages we want a heading left aligned on large displays and center aligned on small-medium displays.

PR #106, opened 2022-03-15
Create responsive text alignment classes

nienn:
Required for:
* #103 
* #104 
* #105

### DIFF
--- a/src/sui/themes/rtd-site/globals/site.overrides
+++ b/src/sui/themes/rtd-site/globals/site.overrides
@@ -55,6 +55,15 @@ body {
 /*-------------------
 Responsive alignment
 --------------------*/
+/* 
+    Usage:
+    - Add classes "text align <alignment> <viewport [viewport] [viewport] [viewport]>" to an element to selectively align its text on the specified viewport(s).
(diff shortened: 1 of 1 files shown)


src/sui/themes/rtd-site/globals/site.overrides
```diff
@@ -52,6 +52,75 @@ body {
   color: @black;
 }
 
+/*-------------------
+Responsive alignment
+--------------------*/
```
agjohnson, line 57: This would be a great place to document what normal usage of these classes looks like. Just a brief example would help those not familiar with normal alignment rules/etc.

agjohnson (approved):
Very neat change! This compliments the SUI built-ins nicely

src/sui/themes/rtd-site/globals/site.overrides
```diff
@@ -52,6 +52,75 @@ body {
...
+
+/* Mobile Only */
+@media only screen and (max-width: @largestMobileScreen) {
+  .text.center.mobile {
```
nienn, line 61:
```suggestion
  .text.align.center.mobile {
```

PR #108, opened 2022-03-15
Improve home page callout responsive behaviour

nienn:
Hides the callout image and centers the text for tablet and bellow viewports.

Closes: #103 
Requires: #106

Preview on mobile:
Screenshot 2022-03-15 091249

### DIFF
--- a/readthedocs-theme/templates/readthedocs/homepage.html
+++ b/readthedocs-theme/templates/readthedocs/homepage.html
@@ -12,24 +12,24 @@
 
         <div class="row">
 
-          <div class="column seven wide computer nine wide tablet sixteen wide mobile">
-            <h1 class="ui header massive">
+          <div class="column seven wide computer twelve wide tablet sixteen wide mobile">
+            <h1 class="ui header massive center align text tablet mobile">
               <span class="ui header">Push your docs, <br>we'll do the rest</span>
             </h1>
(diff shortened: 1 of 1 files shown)


readthedocs-theme/templates/readthedocs/homepage.html
```diff
@@ -12,24 +12,24 @@
...
-            <div>
+            <div class="center align text tablet mobile">
               <a href="https://docs.readthedocs.io/page/intro/getting-started-with-sphinx.html" class="ui button large secondary gradient fade">Get started!</a>
               <a href="./product.html" class="ui button large secondary basic">Learn more</a>
             </div>
           
           </div>
 
-          <div class="column seven wide computer nine wide tablet sixteen wide mobile">
+          <div class="seven wide computer only column">
```
agjohnson, line 32: I assume you meant this instead:

```suggestion
          <div class="seven wide computer only computer">
```

nienn: No. The classes `computer only` are order sensitive and need to be in that order. Also, what you suggest repeats the word computer and removes column.

agjohnson: Yeah I thought the order was `only computer`.

agjohnson (approved):

Issue #86, opened 2022-02-24
Add error pages


agjohnson:
We need:

- 4xx general page
- 403 page
- 404 page
- 5xx page

## Work

- [ ] If there are no immediate design ideas, just a logo and text is enough.
- [ ] #111

nienn:
I'm moving the cloudfront task to a separate issue, as PR #110 will close this one and the cloudfront settings have to be done independently from the current work, by the Ops team.

PR #110, opened 2022-03-16
Error pages

nienn:
Creates a basic template for the HTTP error pages and adds content for the most common HTTP responses.

Resolves: #86


agjohnson:
Because RTD has a bug:

### Computer

image

### Mobile

image

content/pages/errors/403.html
```diff
@@ -0,0 +1,26 @@
+<html>
```
humitos, line 1: This page seem to be exactly the same as 401. Should we create a symlink instead to avoid having to maintain two HTML files with the same content?

content/pages/errors/5xx.html
```diff
@@ -0,0 +1,19 @@
...
+    <meta name="subtitle" content="Something went wrong" />
+    <meta name="template" content="error" />
+    <meta name="slug" content="errors/5xx" />
+    <meta name="status" content="hidden" />
+    <meta name="icon" content="fa-fire-smoke" />
+  </head>
+  <body>
+{% markdown %}
+
+There was a unexpected application error that prevented us from processing
```
humitos, line 13:
```suggestion
There was an unexpected application error that prevented us from processing
```

humitos (approved):
Looks good to me! How are we going to use these templates _from inside the application_?

content/pages/errors/403.html
agjohnson: They are very slightly different for now, but do serve different purposes. 401 is not authorized as in user is not logged in, 403 is access denied as in wrong permisisons. We could make these two more distinct if we want.

agjohnson: I updated the 401 error with a little bit more copy. I'm not even sure we differentiate between the two ever in the application either though.

Issue #102, opened 2022-03-10
Top navigation is clipped at view ports less than 990px


agjohnson:
990px:

![image](https://user-images.githubusercontent.com/1140183/157761831-02af6fd3-cea2-4416-8f63-1e48f7fa4684.png)

mobile:

![image](https://user-images.githubusercontent.com/1140183/157761914-b07f0e67-25ff-45de-b443-ebaa6b1464c9.png)

PR #112, opened 2022-03-17
Make #topnav responsive

nienn:
Stackable menu, relying on the default SUI options:

Screenshot 2022-03-17 093120

* Resolves: #102

### DIFF
--- a/readthedocs-theme/templates/includes/footer.html
+++ b/readthedocs-theme/templates/includes/footer.html
@@ -49,6 +49,7 @@
           <h4 class="ui sub header">Business Info</h4>
 
           <div class="ui list">
+            <a class="item" href="{{ SITEURL }}/company.html">Company</a>
             <a class="item" href="https://www.ethicalads.io/advertisers/?ref=rtd">Advertise with Us</a>
             <a class="item" href="https://readthedocs.com">Read the Docs for Business</a>
             <a class="item" href="https://docs.readthedocs.io/page/privacy-policy.html">Privacy Policy</a>
--- a/readthedocs-theme/templates/includes/topnav.html
+++ b/readthedocs-theme/templates/includes/topnav.html
@@ -7,11 +7,19 @@
 
     <a class="item{% if page and page.slug == 'product' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/product.html">Product</a>
     <a class="item{% if page and page.slug == 'pricing' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/pricing.html">Pricing</a>
-    <a class="item{% if page and page.slug == 'company' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/company.html">Company</a>
-    <a class="item" href="https://docs.readthedocs.io">Resources</a>
 
     <div class="right menu">
-      <a class="item" href="https://readthedocs.org/support">Support</a>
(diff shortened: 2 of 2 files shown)


nienn:
Yes, as we talked before, SUI doesn't have great responsive options. There's no option for the `stackable menu` over the 992px breakpoint for example. Although this could be easily done with some custom rules.

SUI `stackable menu` also does not deal well with dropdown menu. See https://fomantic-ui.com/collections/menu.html#stackable

This is how a dropdown menu appears on mobile:

Screenshot 2022-03-21 091301

The appearance is not the biggest problem however. The dropdown is designed to be toggled on `houver`. It disappears as soon as the `hover` event stops.

Correcting this behaviour should require even more custom code than styling the 991px breakpoint.

Also, is it a good idea to derive the main site nav and user experience from the framework limitations?

agjohnson:
Yeah, I wouldn't combine stackable menus with dropdowns. The docs state that stackable menus are only really helpful for simple menus.

The correct way to do what you are trying to do would be with a sidebar menu:
https://semantic-ui.com/modules/sidebar.html

But that is rather complex for our needs, we don't get much mobile traffic.

I still lean towards my first suggestion, reducing what is in the top navigation. At very least, the link for the company page doesn't seem important enough to users that it should be located at the top of the page.

It looks like we need one more link removed to fit narrow displays. The resources link is one to reconsider, at very least for narrow displays. I think for desktop displays, that should be a dropdown with more explicit menu items:

- Help
  - Support
  - Docs
  - Getting started
  - Configuration file
  - Status

And a third option is two separate menus, one for desktop display, one for tablet/narrow display -- this is two separate list of links, or perhaps `desktop only` on some of the menu items. On narrow displays, we don't use dropdowns.

And perhaps a fourth option is the right menu drops down to it's own row on narrow displays.

Anyways, there are a number of options here I feel, many native to SUI. I'm not too particular on which direction we go, so feel free to take the easiest path here. I don't think the solution needs to be too complicated or custom.

nienn:
> At very least, the link for the company page doesn't seem important enough to users that it should be located at the top of the page.

> It looks like we need one more link removed to fit narrow displays. The resources link is one to reconsider, at very least for narrow displays.

Summarising the work to be done here: 
- [ ] remove the links to "**company**" and "**resources**" from the top menu
- [ ] add the following dropdown, but only for desktop:
```
* Help
  * Support
  * Docs
  * Getting started
  * Configuration file
  * Status
```

Since its on the dropdown for desktop, I'm guessing the "**support**" link should be removed too? 

> The correct way to do what you are trying to do would be with a sidebar menu.

I never suggested a sidebar menu. I don't think it fits our needs here in any way.

nienn:
Last commit:

* The links to "**company**" and "**resources**" were removed from the top menu.
* I added the "**company**" link to the footer.
* The  "**help**" link was added with the submenu in a dropdown:
  <br>
  ```
  * Support
  * Docs
  * Getting started
  * Configuration file
  * Status
  ```
There are extra files and scripts needed for the SUI JS to work properly, but I'm guessing that's duplicate work from the ext-theme project, so it should probably be configured after the projects are merged.

For now, the menu looks like this on different viewports:

_Computer_

computer

_Tablet_

tablet

_Mobile_

mobile

agjohnson:
I opened #117 to address the JS bundle. This looks good in theory so will just get it merged for now.

PR #118, opened 2022-04-06
Try build jobs

agjohnson:
I was mostly curious about what this would look like, but it was rather easy to accomplish.

This enables building with Pelican and uses a dummy Sphinx configuration to mostly skip the Sphinx build step. The one hurdle was knowing you needed to put the files in a certain location to get them cloned to RTD storage.

For this use case, build.jobs + a generic builder that skipped Sphinx and collected my build files would be enough -- so probably no metadata.yaml or build.commands required

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,18 @@
+version: 2
+
+sphinx:
+  configuration: "./conf.py"
+
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.8"
+    nodejs: "14"
+  jobs:
+    pre_build:
+      - pip install -r requirements.txt
+      - npm ci
+      - npm run build-html
+    post_build:
+      - rm -rfv ../../artifacts/${READTHEDOCS_VERSION}/sphinx
+      - mv -v ./output/ ../../artifacts/${READTHEDOCS_VERSION}/sphinx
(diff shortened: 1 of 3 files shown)


.readthedocs.yml
```diff
@@ -0,0 +1,18 @@
...
+    python: "3.8"
+    nodejs: "14"
+  jobs:
+    pre_build:
+      - pip install -r requirements.txt
+      - npm ci
+      - npm run build-html
+    post_build:
+      - rm -rfv ../../artifacts/${READTHEDOCS_VERSION}/sphinx
+      - mv -v ./output/ ../../artifacts/${READTHEDOCS_VERSION}/sphinx
```
humitos, line 18: Haha! Nice hack to support building with pelican 😛

ericholscher: Yea, I figured there would be a blog post about how to support other tools like this in <3 months after we promote this.

agjohnson:
I put this PR up mostly to show @humitos I was doing evil things with his new creation. I'm excited to just roll with this for now though, I'll see how it goes! :upside_down_face:

ericholscher:
@agjohnson does this mean we can host our landing pages on RTD in prod?

agjohnson:
I thought about that, yeah. I'll give it a whirl and see if there is some set up that makes sense here.

PR #138, opened 2022-07-27
Add draft marketing example page

### LINKED ISSUE #121: Add base templates to work from
We wanted to build a few underlying templates to work from, something to cut out all of the redundancy in building html templates and push the content management to rst/md.

This looks like a good place to use a jinja2 rst/md plugin, so that we can populate jinja blocks in an underlying template.

Some layouts that we should have:

- Single column with leading, full bleed hero section at the top. Useful for prominent marketing pages
- Page with leading image, similar to an article but for non blog content
- Page with vertical sections that are split up
- Some form of two column layout, if this can be a generalized base template. Perhaps more likely an include though.
- TBD

Also, likely it might be wise to break out includes:

- Probably two column layout, when we want side by side content like supporting images/etc.
- Inverted section for breaking up long content
- TBD

It might just be worth implementing these all as macros or includes or something. https://github.com/pelican-plugins/jinja2content probably helps us.

agjohnson:
This is a demo page that has a number of blocks that can be copy pasted
for a reasonable base to work from.

### Computer

image

### Tablet

image

### Mobile

image

Closes #121

### DIFF
--- a/content/pages/examples/marketing.html
+++ b/content/pages/examples/marketing.html
@@ -1,0 +1,413 @@
+<html>
+  <head>
+    <title>Example marketing page</title>
+    <meta name="template" content="page" />
+    <meta name="slug" content="examples/marketing" />
+    <meta name="status" content="draft" />
+    <meta name="description" content="An example marketing page" />
+  </head>
+  <body>
+
+{#
+  View this page at http://localhost:8080/drafts/pages/examples/marketing.html
+#}
+
+{% import "partials/basic.html" as basic %}
+{% import "partials/marketing.html" as marketing %}
+{% import "partials/technical.html" as technical %}
+
+<section>
+  {% call basic.header_image(
+    header="Here's an example of a large image header",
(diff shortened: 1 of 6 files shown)


humitos (approved):
I have opposite opinions at the same time here 😄 

I like the idea of having this kind of components that we can re-use across multiple pages. However, at first sight I found them hard to read/use because this is pretty custom --and easy to get confused with the `call` macro, the arguments that it receives and its content.

I suppose that we will get used to this sooner than later and we will start taking advantage of them immediately, tho. I'll figure it out when I start creating landing pages, probably 😄 

Cool documentation for each of the macros 💯

humitos:
> For review purposes, are there any other content blocks that you'd like to see illustrated?

I think we are fine to merge this and start playing with them together with Eric. If we find we need something else, we will let you know and ask for some frontend help 😄

agjohnson:
Yes agreed. It's wise to not spend too much time fiddling with the HTML and just hand this off if you need something more specific.

PR #140, opened 2022-07-28
Add opengraph include

### LINKED ISSUE #139: Add opengraph meta data to base template
The templates are currently missing opengraph meta data. For now, there are a few basic graphics that can be used for the image, but this should be configurable in some fashion later.

agjohnson:
Closes #139


humitos (commented):
Looks good! We were using this logo style for opengraph in the blog and other places --which looks better to me: https://github.com/readthedocs/guidelines/blob/main/assets/logo-wordmark-vertical-compact-dark.png

agjohnson:
Yeah, I didn't have a strong preference between the two. I'm not a huge fan of either, really, and think we can do better with something more opinionated, like:

https://github.com/readthedocs/guidelines/issues/1#issuecomment-885186136

The website uses much more color now, so there's less conflict using colorful supplementary branding too :rainbow:

agjohnson:
Aye, agreed. I haven't had a chance to update this yet, but will update the graphic and open an issue to track making a new graphic.

PR #142, opened 2022-08-02
Dockerize development environment

benjaoming:
* [ ] Install all dependencies runtime
* [ ] Install all dependencies in Docker container
* [ ] Install all dependencies in Docker container, but update runtime
* [ ] Invoke dev server in entrypoint
* [ ] Write good docs

I noticed that the Poetry configuration refuses to install on Python 3.9 so I updated the docs to reflect this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,14 +33,25 @@ You will need to specify the ``FONTAWESOME_NPM_AUTH_TOKEN`` environment variable
 before installing Node dependenvies. You should be using ``direnv`` locally for
 secret management and encryption. Ask for this token, it's private.
 
-You need to be using Node<=14 and at least Python 3.6+, but Python 3.10 is
-preferred. Using ``asdf`` is recommended here. With both configured, you can
-install all the dependencies with:
+You need to be using Node<=14 and Python 3.10.
+Using ``asdf`` is a great option for adding Python 3.10 to a system that doesn't have it. 
+h both configured, you can install all the dependencies with:
 
 .. code-block:: console
 
     $ poetry install
     $ npm install
+
+Running in Docker
+~~~~~~~~~~~~~~~~~
+
+If you have Docker with docker-compose available, you can run a local development server like this:
+
+.. code-block:: console
+
+    $ # invoke direnv to add secret environment
+    $ docker-compose build  # builds the server
+    $ docker-compose up 
 
 Local development and authoring
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,18 @@
+version: '3'
+
+volumes:
+  build:
+
+services:
+
+  server:
+    # Image used for all the other services (proxito, web, celery, build)
+    build:
+      context: ${PWD}
+      dockerfile: ./Dockerfile
+    environment:
+      - FONTAWESOME_NPM_AUTH_TOKEN=${FONTAWESOME_NPM_AUTH_TOKEN}
+    ports:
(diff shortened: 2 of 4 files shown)


agjohnson:
If you are concerned about security locally, this might be a place to just use a one-off server instead of reworking the workflow in this repository. A private container maintains all of the existing workflow, and doesn't add the complexity of another layer on top of local dependencies for everyone.

ericholscher:
@benjaoming I tried to get this going, but it wasn't working for me initiall. I don't know if you can actually do 2 `FROM` statements like that.

I have a slight preference for Ubuntu over Alpine to match the rest of our stack, but would be good to get some standard env setup and working in Docker 👍

benjaoming:
@ericholscher After abandoning this work, I got much further with a similar approach over in sphinx_rtd_theme: https://github.com/readthedocs/sphinx_rtd_theme/pull/1319

My approach has been to:

a) Copy stuff during build such that as many dependencies as possible can be built into the image
2) Mount in everything via bound mounts, such that the runtime environment can quickly reload when files are edited
3) Copy out built assets with `docker cp` (rather than building in the bind mount because of file permissions)
4) Create a make file with convenient shortcuts (also skipped docker-compose)
5) I did also have success running the same Docker environment in Circle CI, even running it through said `make` shortcuts for 1:1 consistency. That enables us to both reuse _and_ test the dev setup.

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    # Image used for all the other services (proxito, web, celery, build)
+    build:
+      context: ${PWD}
+      dockerfile: ./Dockerfile
+    environment:
+      - FONTAWESOME_NPM_AUTH_TOKEN=${FONTAWESOME_NPM_AUTH_TOKEN}
+    ports:
+      - "8000:8000"
+    volumes:
+      - ${PWD}/:/project
```
benjaoming, line 18: There is just one container, so we could do two things:

* Maintain YAML setup in docker-compose, keeping in mind the extra dependency (docker-compose is a dependency!)
* Convert to a `docker run` command, store it in a Makefile.

README.rst
```diff
@@ -33,15 +33,26 @@ You will need to specify the ``FONTAWESOME_NPM_AUTH_TOKEN`` environment variable
...
+Running in Docker
+~~~~~~~~~~~~~~~~~
+
+If you have Docker with docker-compose available, you can run a local development server like this:
+
+.. code-block:: console
+
+    $ # invoke direnv to add secret environment
+    $ docker-compose build  # builds the server
+    $ docker-compose up
```
benjaoming, line 54: This is the part that I contemplate to replace with a Makefile, in case it can get rid of docker-compose. However, if docker-compose keeps being so simple, we might as well just keep using it.

agjohnson: Yeah, I was confused why this wasn't using docker compose. It's my understanding that docker compose is a good fit for use cases like these, and seems to give us all of the definition to mounts and ports.

docker-compose.yml
agjohnson: I think docker compose is the more expected option, no?

benjaoming: Yes, well, you can argue in both directions, but we'll opt for the docker-compose strategy :+1:

benjaoming:
We'll be opting for docker-compose.

ericholscher:
> We'll be opting for docker-compose.

Yea, I think this makes sense. @benjaoming if you have an updated image that would be 💯, so I can test against it.

ericholscher:
Think we can close this one for now, since we aren't prioritizing this work yet.

PR #150, opened 2022-08-18
Add a `stretched row` for equal height columns in the signup modal

agjohnson:
Just noticed the columns weren't the same height, which wasn't intended.
Adding the stretched row is what makes the segments fill the full
row/table height.

----
:books: Documentation preview :books:: https://read-the-docs-site-community--150.com.readthedocs.build/en/150/

### DIFF
--- a/content/partials/homepage.html
+++ b/content/partials/homepage.html
@@ -13,82 +13,85 @@
       </div>
 
       <div class="ui stacking centered grid">
+        <div class="stretched row">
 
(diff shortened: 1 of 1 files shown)


ericholscher (approved):
Would be useful to have a before/after screenshot here, to help with review.

agjohnson:
GitHub was failing to attach files yesterday, unfortunately.

agjohnson:
This slight bug was only at tablet viewport sizes, when a line wraps

# Before



# After

PR #155, opened 2022-08-30
plausible track clicks

ericholscher:
- Track Signup & Login clicks on plausible
- Use slug events to keep it simple.

### DIFF
--- a/readthedocs_theme/templates/includes/plausible-events.html
+++ b/readthedocs_theme/templates/includes/plausible-events.html
@@ -1,0 +1,33 @@
+{# 
+    Enable us to track clicks just by setting a `data-analytics` attribute on links.
+    Docs: https://plausible.io/docs/custom-event-goals#code-example-for-tracking-link-clicks
+#}
+<script>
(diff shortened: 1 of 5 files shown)


readthedocs_theme/templates/includes/plausible-events.html
```diff
@@ -0,0 +1,33 @@
...
+        if (!link.target) {
+            if (!(event.ctrlKey || event.metaKey || event.shiftKey) && click) {
+                setTimeout(function () {
+                    location.href = link.href;
+                }, 150);
+                event.preventDefault();
+            }
+        }
+    }
+  </script>
```
agjohnson, line 33: We should move this to the standard site JS source, instead of using inline JS. I can move it real quick.

ericholscher: Cool -- I thought it was maybe nicer to keep on it's own for clarity, but happy with whatever approach makes sense.

agjohnson: I addressed some issues in #156 and tested that we're receiving events in Plausible

PR #156, opened 2022-08-31
Move Plausible integration to our standard library

agjohnson:
With the addition of JS logic to make Plausible work, it should be
defined alongside our other JS sources. Generally speaking, this is the
pattern we're taking with application and website JS. The reasons for
this are to remove the need for inline JS for security reasons, to push
dependencies into a place they can be tracked, to maintain all JS code
using linting/testing/etc tools, and to bundle everything up into a
single JS artifact.

A few key differences between the JS here and the JS in #155:

* This JS works for more than just anchor elements. Using SUI, we'll want to
  track on buttons, divs, etc, in addition to anchor elements
* Use the Plausible library they publish on NPM instead of the include
  script inline
* Rely on a callback to redirect the browser
* Some cleaner code around conditionals

----
:books: Documentation preview :books:: https://read-the-docs-site-community--156.com.readthedocs.build/en/156/


src/js/site.js
```diff
@@ -86,5 +90,75 @@ jquery.fn.sui_module = function () {
...
+ * cases, this should be a link element, however in the case of other UI
+ * components, it may be a ``<div>`` or ``<button>``:
+ *
+ *     <button class="ui button" data-analytics="some-event-id">Something</button>
+ *
+ * In the case of a link with a ``href`` attribute, the link will continue
+ * redirecting after the callback from Plausible fires off.
+ */
+jquery.fn.plausible = function () {
+  let plausible_settings = { domain: "about.readthedocs.org" };
```
humitos, line 107:
```suggestion
  let plausible_settings = { domain: "about.readthedocs.com" };
```

ericholscher: We need to update this in plausible. I can do that today.

ericholscher (approved):
This looks good 👍

src/js/site.js
```diff
@@ -86,5 +90,75 @@ jquery.fn.sui_module = function () {
...
+      const is_click = event.type == "click";
+      const is_link_click =
+        is_link &&
+        is_click &&
+        !elem.target &&
+        !(event.ctrlKey || event.metaKey || event.shiftKey);
+
+      if (is_middle_click || is_click) {
+        function redirect_link() {
+          if (is_link_click && elem.href && elem.href != "#") {
```
agjohnson, line 130: I had a random thought, and this logic might not be correct for anchor links. Let me double check this really quick.

agjohnson: Nevermind, works as expected! My concern was `href="#foo"` would not have the default event propagation, but it does indeed work.

src/js/site.js
ericholscher: Updated.

agjohnson:
Looks like this failed the build check, I'll open a separate PR

PR #159, opened 2022-08-31
Add tracking to the bottom CTA on the homepage

agjohnson:
I still dislike this block, but perhaps a link is at least helpful for now.

- Requires #160, merge that first

----
:books: Documentation preview :books:: https://read-the-docs-site-community--159.com.readthedocs.build/en/159/


agjohnson:
This might be a good block to replace with a call out to our example projects in the future. Ideally "Get started", in the context of "get online in 5 minutes", should do more than link to our sign up page. I'm +1 on reframing this CTA as something more helpful though, I'm not particularly jazzed on the framing on around speed.

agjohnson:
#161 adds a modal, which seems fine. We can come back to this block and refocus the CTA later.

PR #166, opened 2022-11-10
Rework pricing page

agjohnson:
This fixes up a number of content issues with the pricing page:

- Having two rows of plans with competing labels is confusing. It's not guiding
  the user towards "Read the Docs for Business" or "Base plans"
- The feature comparison is still really hard

Changes are:

- Add a tabbed segment for community and commercial plans, so we can limit what
  the reader is looking at
- Keep all plans listed above the fold
- Adds the big feature checklist we've wanted, so we can be less verbose above
  the fold
- Some technical fiddly bits to support tabbed segments that can be linked to
- Some technical CSS bits to make tablet display less bad

Last steps:

- [x] Go back over feature list icons and copy, and supporting text copy
- [x] Add analytics back in

----
:books: Documentation preview :books:: https://read-the-docs-site-community--166.com.readthedocs.build/


humitos:
It's looking great! 💪🏼

ericholscher:
I love the feature table! Been wanting that for a long time.

agjohnson:
I was using the (archived?) version here: https://github.com/readthedocs/policy/blob/main/pricing.csv

ericholscher:
Ah yea, I think is an export of it :)

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
-              </div>
+  {% block pricing_community_intro %}
+    <div class="ui padded container">
+      <div class="ui clearing inverted teal segment">
+        <div class="ui stackable grid">
 
-            </div>
+          <div class="sixteen wide tablet twelve wide computer middle aligned column">
+            <i class="fad fa-trees icon"></i>
+            Host <b>open source and community projects</b> for free with Read the Docs Community.
```
ericholscher, line 22: I do wonder about putting this callout lower on the page. Perhaps above the feature checklists?

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
+    <div class="ui horizontally fitted basic active tab segment" data-tab="business">
+      <div class="ui centered three column tablet stackable grid">
+        <div class="stretched row">
+
+          {% call pricing.plan(
+            name="Basic plan",
+            price="$50",
+            color="violet",
+          ) %}
+            <div class="ui list relaxed divided">
```
ericholscher, line 63: We lost the little copy that we have about who the plan is good for (eg `For public-facing docs on your domain`). The feature listing mostly covers this, but I do think these short snippets are useful to add back to contextualize the plans.

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
-              <span class="right floated"><i>limited time</i></span>
+          {% endblock pricing_opensource %}
+
+          {% call pricing.plan(
+            name="Gold membership",
+            price="$5",
+            color="teal",
+          ) %}
+            <div class="ui list relaxed divided">
+              {{ pricing.plan_feature("Public versions", "Publicly accessible projects only", "fa-shield-keyhole") }}
```
ericholscher, line 139: Not sure what this means?

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
+              <div class="ui padded inverted teal segment">
+                <h3 class="ui small header">
+                  <i class="fad fa-circle-question icon"></i>
+                  <span class="content">
+                    Which platform?
+                  </span>
+                </h3>
+
+                <p>
+                  Read the Docs Community is intended for <b>open source</b> and <b>free</b> software,
```
ericholscher, line 155: I assume we're using free as in beer here? We should probably be more explicit. We used `open source and community` elsewhere on the page, perhaps do the same here?

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
+  </div>
+</section>
+
+{# Compare plan features with a table #}
+<section id="compare">
+  <div class="ui padded container">
+
+    {% call pricing.feature_table() %}
+      {{ pricing.feature_row("Team management", "Control organization access with team permissions", "fa-users", on_community=False) }}
+      {{ pricing.feature_row("Email support", "Get help with your projects and documentation", "fa-message-question", None, "2 business days", "2 business days", "1 business day") }}
```
ericholscher, line 196: We do actually offer pretty good email support on community, but probably better to underpromise here.  🤷

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
+      {{ pricing.feature_row("Pull request preview", "Preview changes in review", "fa-code-compare") }}
+
+      {{ pricing.header_row("Documentation building", "fa-gears") }}
+      {{ pricing.feature_row("Automatic building", "Build every commit", "fa-gears") }}
+      {{ pricing.feature_row("Build concurrency", "Concurrent build processes", "fa-rectangle-vertical-history fa-swap-opacity", 2, 2, 4, 6) }}
+      {{ pricing.feature_row("Notifications", "Receive notifications build events", "fa-envelope-dot") }}
+      {{ pricing.feature_row("Environment variables", "Inject secrets into builds", "fa-lock-keyhole") }}
+      {{ pricing.feature_row("Automation rules", "Automatically manage versions", "fa-robot") }}
+
+      {{ pricing.header_row("Documentation hosting", "fa-browser") }}
```
ericholscher, line 212: This checklist makes it really obvious that our Pro plan isn't differentiated enough 🙃

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
+      {{ pricing.feature_row("Documentation versioning", "Host multiple versions of your project", "fa-layer-group") }}
+      {{ pricing.feature_row("Documentation localization", "Host your project localized in multiple locales", "fa-language") }}
+      {{ pricing.feature_row("Reader authentication", "Require access to view your documentation", "fa-shield-keyhole", on_community=False) }}
+      {{ pricing.feature_row("Single sign-on", "Sign on with using external provider", "fa-right-to-bracket", on_community=None, on_basic="GitHub and GitLab", on_advanced="GitHub and GitLab", on_pro="GitHub, GitLab, and Google") }}
```
ericholscher, line 216: I feel like we're not selling the Pro plan enough here. I think it probably makes sense to split Google Auth out into its own line where only Pro has it.

ericholscher (approved):
This is a great improvement! It really shows two big things to me:

* We need to differentiate the Pro plan
* We need to differentiate the Gold plan

We already knew both of these things, but it's really obvious when listed in a big grid 🙃

content/partials/pricing.html
```diff
@@ -0,0 +1,171 @@
+{#
+  It's not yet clear if the community plan features on the feature table help
+  explain the difference between community/commercial, or if it makes it less
+  obvious that community is only for open source projects.
```
ericholscher, line 4: It would be nice if we could toggle them on & off, and default them off, but you can turn it on with a toggle.

content/pages/pricing.html
```diff
@@ -2,389 +2,263 @@
...
+      {{ pricing.feature_row("Request redirects", "Redirect requests for moved and deleted content", "fa-diamond-turn-right") }}
+      {{ pricing.feature_row("Sitemap generation", "Automatically update your project's sitemap", "fa-sitemap") }}
+      {{ pricing.feature_row("Customizable error pages", "Manage your own HTTP error pages", "fa-palette") }}
+      {{ pricing.feature_row("Status badges", "Embeddable badges showing your project's status", "fa-badge-check") }}
+      {{ pricing.feature_row("Google analytics integration", "Track visits to your project", "fa-chart-line") }}
+      {{ pricing.feature_row("Pageview analytics", "Collect statistics on visits to your projects", "fa-chart-line", on_community="30 days", on_basic="30 days", on_advanced="30 days", on_pro="90 days") }}
+      {{ pricing.feature_row("Search analytics", "Collect statistics on search queries for your projects", "fa-chart-line", on_community="30 days", on_basic="30 days", on_advanced="30 days", on_pro="90 days") }}
+      {{ pricing.feature_row("Content embedding", "Embed your documentation in outside sources", "fa-object-subtract", on_basic=False) }}
```
ericholscher, line 231: I do wonder about putting all the features that are differentiated at the top? That way we have a nice visual for what is included on each plan, and then all the joint features below? Not 100% sure this is better, but I find it hard to easily scan what is diff between plans.

content/pages/pricing.html
agjohnson: I also thought about highlighting these table cells differently, to provide more contrast. I can try that first, I felt that multiple rows defeated the purpose of building a table.

content/partials/pricing.html
agjohnson: Mm, that could be a good option. Okay to leave this one for later? I was going to wait to see on this and maybe play more in another pass. I could see us lumping the Gold plan in the feature table too, but avoided making this table too confusing for now.

content/pages/pricing.html
agjohnson: There wasn't a clear way to organize this list, because even as we get lower on the features table, the community plan still has features that the basic plan doesn't, so it's hard to organize by plan like what you're describing.

I did find it easy to scan by splitting up by concern -- build, hosting, etc -- but this doesn't help highlight any particular plan necessarily.

content/pages/pricing.html
agjohnson: Yuup. And we knew that of course, but this view does make it more obvious

content/pages/pricing.html
agjohnson: Yeah, i was not entirely sure how to frame this one in particular. We support the community platform of course, but support isn't necessarily part of the community product. It is definitely part of the commercial product though.

I thought about just making this one a check mark, and the commercial plans would state the "2 business days"/etc in this row. It's ambiguous what community support is in this case, but at least looks like we're responsible for something :smile: 

image

content/pages/pricing.html
agjohnson: Good point, yeah I'll update to match

content/partials/pricing.html
ericholscher: Yea, definitely not a blocker.

content/pages/pricing.html
agjohnson: Ah, oops yes, this was a placeholder. I wanted another feature in the list to fill it out. I should change this to mention documentation is ad-free as a reader.

content/pages/pricing.html
agjohnson: Yeah, I tried to fit these in but they made these lists more cluttered. I did like the addition in theory though, but perhaps I'll put testing these out again back on my later list at #167

content/pages/pricing.html
agjohnson: image

content/pages/pricing.html
agjohnson: Yeah, I'm leaning towards highlight:

image

content/partials/pricing.html
agjohnson: I put this one on my later list on #167. We can return to this and polish the UX on the community plans more, I think I'd like some time for this to simmer and see if we're confusing folks more or less with this listing.

content/pages/pricing.html
ericholscher: Yea, that works well 👍

content/pages/pricing.html
agjohnson: Hrm, good question, both look good. Here's above the feature list:

image

content/pages/pricing.html
ericholscher: Yea, I think splitting by section is good, I was mostly suggesting the sorting within the section. Definitely not a fully baked thought though.

agjohnson: I put this one on #167 as well, but where I stand right now is that major features should be differentiated in the top plan section instead. The feature table should be more organized -- grouped by function and context makes the list easier to scan.

content/pages/pricing.html
agjohnson: I actually do like the check, it's a rather simple solution. If we get questions here we can elaborate, but I actually think the check works well for this purpose, and differentiates commercial nicely.

content/pages/pricing.html
agjohnson: I have slight preference to having it first, but dropped it down for now. I'll note on #167 but I think it does look good between plans/features, so not a big issue at all either way.

agjohnson:
A put a few of these on the later polish list, #167, and we can revisit them later with some more usage on this page.

content/pages/pricing.html
ericholscher: Yea, it looks a bit redundant with that placement right above the feature list 🤔

ericholscher:
Happy to ship this 👍

content/pages/pricing.html
agjohnson: Yeah, it's an odd piece. I'm happy with it as it is here and #167 can revisit some fixes perhaps.

PR #170, opened 2022-11-15
Fix top nav width at computer and tablet

agjohnson:
The topnav was tuned for tablet display, where the logo column is one
column wider. On desktop viewport sizes, the topnav showed one column
too narrow.

Tablet:
image

And desktop viewport is now full width.

Before:
image

After:
image

----
:books: Documentation preview :books:: https://read-the-docs-site-community--170.com.readthedocs.build/en/170/


humitos (approved):
I'm not able to notice any difference between the current code and this PR using Firefox and the different viewports emulation it has. I suppose this is fine 🤔

PR #172, opened 2022-11-19
Move common SUI theme to it's own repo

agjohnson:
This moves shared and common styles between our website and application
to it's own shared package. Both the website and application can install
from this repository, via NPM Git install.

This is pinned at a commit, but will eventually be tags/releases.

The biggest change here is that this drops the need for FontAwesome Pro
in this repository. Instead, we use our own hosted kit at FontAwesome.
This provides a big performance benefit, and avoids having to manually
optimzie the FA webfonts or package them in our own bundles.

Refs https://github.com/readthedocs/ext-theme/issues/110

----
:books: Documentation preview :books:: https://read-the-docs-site-community--172.com.readthedocs.build/en/172/


ericholscher (approved):
Screenshot 2022-12-07 at 10 57 23 AM
😍

PR #173, opened 2022-11-19
Use asdf for system packages

agjohnson:
We've talked about checking in the `asdf` configuration file for explicit
version management of system dependencies in the past. This is a lightweight
version of that that gives an explicit way to get started. Using `asdf` locally
is out of the scope of documentation here, it's a system level dependency.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,13 +29,22 @@ Setting up your environment
 This package uses Poetry to manage Python dependencies and CSS and JS assets are
 built using a number of Node dependencies.
 
-You will need to specify the ``FONTAWESOME_NPM_AUTH_TOKEN`` environment variable
-before installing Node dependenvies. You should be using ``direnv`` locally for
-secret management and encryption. Ask for this token, it's private.
+You will need the following requirements:
 
-You need to be using Node<=14 and at least Python 3.6+, but Python 3.10 is
-preferred. Using ``asdf`` is recommended here. With both configured, you can
-install all the dependencies with:
+* Node.js version ``14``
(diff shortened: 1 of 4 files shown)


agjohnson:
When we talked about this one, @humitos raised issues about `.tool-versions` changing suddenly. This could be a problem, though with a repository configuration that uses NPM and Poetry, the system dependencies can change underneath without affecting the installed dependencies.

README.rst
```diff
@@ -29,13 +29,23 @@ Setting up your environment
 This package uses Poetry to manage Python dependencies and CSS and JS assets are
 built using a number of Node dependencies.
 
-You will need to specify the ``FONTAWESOME_NPM_AUTH_TOKEN`` environment variable
-before installing Node dependenvies. You should be using ``direnv`` locally for
-secret management and encryption. Ask for this token, it's private.
+You will need the following requirements:
```
agjohnson, line 32: This change is related to #172, the token is not required because we're using a hosted kit.

humitos (approved):
Looks good to me.

It would be good if `.tool-versions` has a way to specify `3.10.x` so, any patch version that you have installed works. Otherwise, we will end up with multiple 3.10 versions installed locally. It's not a problem, of course, just a nice to have 😄

agjohnson:
Yeah, I wish there was some concept of version selection like that in asdf. There are a few issues, like https://github.com/asdf-vm/asdf/issues/352, that discuss some maybe options, but seems this feature was specifically ruled out.

At least with poetry abstracting the Python version a little, it's mostly just wasting the space for each Python core installation, but even a bare installation is about 250MB per install :/

PR #174, opened 2022-11-25
Update About page -> How it works

ericholscher:
This makes it a bit more obvious what the page is about,
and isn't confused with the `About us` page.

Already put in a redirect:

Screenshot 2022-11-25 at 9 07 42 AM

----
:books: Documentation preview :books:: https://read-the-docs-site-community--174.com.readthedocs.build/en/174/


readthedocs_theme/templates/includes/topnav.html
```diff
@@ -1,5 +1,5 @@
 {% macro menu_main() %}
-  <a class="item{% if page and page.slug == 'about' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/about/">About</a>
+  <a class="item{% if page and page.slug == 'about' %} active" aria-current="page{% endif %}" href="{{ SITEURL }}/how-it-works/">How it works</a>
```
agjohnson, line 2: "How it Works" doesn't look good enough for me to want to break the pattern of single word entries in the top navigation. The website has a few entries right now, but I'd like more, and the application will have more top navigation entries.

agjohnson (commented):
I'm more impartial on the name of the page. "About" seems fine as the target page is serving two purposes -- feature list and explaining how RTD works.

I'd suggest not using "How it Works" and instead call the page "Overview", "Features" (though this page is more than features), or "Product" (though I hate this pattern in general for being incredibly unclear). I'd probably lean "Overview".

I proposed putting this under a dropdown, so we can be more wordy with navigation entries. We likely want to link other pages as well, which I why I haven't bothered changing this page title yet.

ericholscher:
@agjohnson I'm confused, since you suggested the exact opposite on Slack: https://readthedocs.slack.com/archives/C1VM2ENUB/p1668462658520369 -- I originally suggested Product Overview, which I think is the right framing.

agjohnson:
My suggestion in that conversation was to put verbose options in a dropdown menu, which is in line with my suggestion above. Guessing this is missing from the context, as I thought you agreed with my suggestion for the dropdown in that conversation :shrug:

> so are you +1 with Overview and we can iterate on a larger menu later?

I mean, I don't share the priority on renaming the `about` page, so no. I would replace the top navigation with a dropdown first, which is what I described in that conversation:

https://readthedocs.slack.com/archives/C1VM2ENUB/p1668462728477219

Thinking farther ahead, "Overview" is okay title for the `about` page, but I wouldn't reuse "Overview" for a dropdown later. So this would be a bad half-step. As a navigation link to a page, "Product" is horribly unclear -- it doesn't describe what to expect at all. But, as a dropdown, "Product" works better than other options, as these page links will be verbose submenu entries inside the dropdown anyways.

So to clarify, I would:

* Not rename the `about` page yet
* Replace the "About" top navigation menu with a "Product" dropdown
* Move the "about" page to "product/how-read-the-docs-works" and add a submenu item in the "Product" dropdown for "How Read the Docs works"
* Finally add necessary redirects

If you feel like managing this change and redirects twice, you can rename "about" to "product". Without the dropdown, this is a lateral move though, and doesn't seem any clearer. So seems like wasted effort.

ericholscher:
OK. I will close this then, and wait for you to do a dropdown 👍

PR #177, opened 2022-12-24
Slight tweak to header menu padding

agjohnson:
On our application, there was a lot of vertical padding on larger
elements, like buttons. This caused uneven height between the
application and website. This puts the padding on the image, which is a
static height, and the rest of the elements align to that instead.

----
:books: Documentation preview :books:: https://read-the-docs-site-community--177.com.readthedocs.build/en/177/


ericholscher:
Should we merge this?

agjohnson:
Yeah, this should be safe to get merged.

PR #181, opened 2023-03-03
Add a small callout alerting users to the new site

ericholscher:
This is just a mockup,
happy to have additional design advice.
I think this could be a "banner" style or similar,
and perhaps only show when we have the `ref=dotcom` header if there's an easy way to do that.

## Preview

Screenshot 2023-03-02 at 7 26 09 PM

----
:books: Documentation preview :books:: https://read-the-docs-website--181.com.readthedocs.build/


humitos (commented):
This is really good! I only made a comment about the style.

readthedocs_theme/templates/base.html
```diff
@@ -61,6 +61,14 @@
 <body>
   {% include 'includes/topnav.html' %}
 
+  <div class="ui center aligned header">
```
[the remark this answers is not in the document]
ericholscher: Yea, I found this, which might work well: https://semantic-ui.com/collections/message.html

Screenshot 2023-03-06 at 6 34 02 AM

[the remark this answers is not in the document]
agjohnson: You can use a `message` or an `inverted segment`, both would be a good fit here.

I'm eventually going to want to tweak the `message` styling for the color variants, but a `teal message` or `primary message` should fit fairly well.

readthedocs_theme/templates/base.html
```diff
@@ -61,6 +61,14 @@
...
+        {# TODO: Replace with live URL #}
+        <a href="https://readthedocs-blog--212.org.readthedocs.build/website-migration/">
+        <i class="fad fa-rocket-launch icon"></i>
+        Welcome to our new website!
```
agjohnson, line 68: I didn't think to click this heading. It would help to find some supporting text/subheading that mentions the blog post or something similar.

ericholscher: Yea, this is just an example. Thoughts on what would work best here in SUI?

ericholscher:
@agjohnson I'd like to get something like this put together for when we migrate .org users to the new website. I know Benjamin was worried about folks being worried about Community going away, so we should try to clarify this as much as possible.

readthedocs_theme/templates/base.html
agjohnson: I mentioned above using `.ui.message`, though I tried this and it took up a lot of screen real estate with an icon, header, and supporting line below. I did a hybrid approach with 1be48d7

agjohnson:
I pushed up an option at 1be48d7

It does move the call out to only on the homepage, as putting it on all pages could alter those pages, and the block isn't dismissable.

The copy could be improved there, but I stuck to using a single line (to avoid altering the page layout above the fold too much. I updated the link as well, but we will want to update that blog post around the same time too.

ericholscher:
This looks like a nice approach! I think this looks good for when we redirect .org, with the content updates you mentioned 💯

agjohnson:
I'll set this to blocking until we get the blog post updated and the link updated here if needed.

ericholscher:
This should now be unblocked with the updates here: https://github.com/readthedocs/blog/pull/236

agjohnson (approved):

PR #182, opened 2023-03-09
Clarify gold pricing

ericholscher:
----
:books: Documentation preview :books:: https://read-the-docs-site-community--182.com.readthedocs.build/

### DIFF
--- a/content/pages/pricing.html
+++ b/content/pages/pricing.html
@@ -111,7 +111,7 @@
 
           {% call pricing.plan(
             name="Gold membership",
-            price="$5",
+            price="$5 and up",
             color="teal",
           ) %}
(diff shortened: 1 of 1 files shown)


agjohnson:
The price looks a little funny with text in that price field:

image

Still not great, but and extra label above is a little better:

image

This is:

```
<div class="statistic">
  <div class="label">Starting at</div>
  <div class="value">$5</div>
  <div class="label">
    <span class="category">per month</span>
  </div>
</div>
```

The macro will need updated with another field

ericholscher:
Yea, not a huge priority, just noted it when talking about it on Twitter. I think probably doing a third level that's "Company" pricing is the right approach thinking more about it.

PR #194, opened 2023-04-19
Page: docs-as-code

humitos:
Initial work to talk about "Docs-as-code concept" and how it's integrated with Read the Docs getting the main features by default thanks to the `readthedocs-client.js`.

The idea is to expose all these main features as
https://github.com/readthedocs/readthedocs-client does by talking about docs-as-code and "all the doctools".

At the bottom there should be a copy&paste examples with one pretty known doctool different than Sphinx and MkDocs (maybe Docusaurus).

📚 Rendered version of **this page**: https://read-the-docs-website--194.com.readthedocs.build/docs-as-code/

----
:books: Documentation preview :books:: https://read-the-docs-website--194.com.readthedocs.build/


ericholscher:
I'd be curious to get this merged just to get it ranking on Google, and we can improve it over time. I'll do a quick once over before we ship it. I'd love a better pattern for showing the 4 "items" on the right are actually clickable headings.

ericholscher:
Actually looking at this more, it seems like it's a "addons" or "documentation features" landing page more than docs as code? It doesn't really talk about docs as code at all. That's not a bad thing -- perhaps we can reframe it to focus on readers?

humitos:
My goal here was not to explain what "Docs-as-code means". I think we can briefly mention it in one sentence or similar. My idea with this page is to attract people looking for "Docs as code hosting" and tell them that we support the workflow they want and we add a lot of good features on top of that automatically.

On this page, I'm assuming that people looking for "docs as code" already know what that means.

humitos:
@ericholscher I updated this page to show the addon features in a nice way. There are some work to do with the copy still, but I think the general idea is OK. I also want to use a Docusaurus example for the images, but I think it will be easier next week after deploying all the addons.

humitos:
> I also want to use a Docusaurus example for the images, but I think it will be easier next week after deploying all the addons.

I updated the PR to use Docusaurus images for the examples 👍🏼 . I think it's looking a lot better now.

ericholscher (approved):
This looks great to me. There's a small bit of additional cleanup I'd like to do later, but I think it's good enough to ship as is.

readthedocs_theme/templates/includes/footer.html
```diff
@@ -30,6 +30,7 @@ <h4 class="ui sub header">Learn more</h4>
               <a class="item" href="/comparisons/github-pages/">Comparison to GitHub Pages</a>
               <a class="item" href="/comparisons/gitbook/">Comparison to GitBook</a>
               <a class="item" href="/comparisons/cloudflare-pages/">Comparison to Cloudflare Pages</a>
+              <a class="item" href="/docs-as-code/">Docs as code</a>
```
ericholscher, line 33: We're definitely piling more into the footer here. I think it's about time to break these out into links in the main content and a comparisons subpage.

content/pages/docs-as-code.html
```diff
@@ -0,0 +1,266 @@
...
+              {%- set icon_4 = "fa-file-plus-minus" -%}
+              <a href="#visual-diff" class="ui small header item">
+                <i class="fad {{ icon_4 }} secondary big icon"></i>
+                <span class="content">
+                  {{ header_4 }}
+                  <span class="sub header">Review PR changes visually.</span>
+                </span>
+              </a>
+
+              {%- set header_5 = "Read the Docs authentic flyout" %}
```
ericholscher, line 101: Not sure what an authentic flyout is? 

```suggestion
              {%- set header_5 = "Read the Docs flyout" %}
```

readthedocs_theme/templates/includes/footer.html
humitos: Yeah. I added it here because it's the easy one, but we need to find a better pattern. What are you doing in EthicalAds for these pages? Where are you linking them from? Do you have other index pages that link these all?

ericholscher: @humitos generally they are linked from inside another page. The "docs as code" page for example could just be linked from copy on our features page talking about this, or something.

content/pages/docs-as-code.html
humitos: Now that you mentioned it, maybe it doesn't make too much sense.

I thought the flyout as something that people already knew about it, but that's a mistake. In fact, I'm not even sure that people will know what a flyout is at all 😅

Maybe it's better to call it in a more generic way like "version selector" or similar.

humitos: Well, the second sentence explains what it is, so that may be enough in this context

humitos: Well, the second sentence explains what it is, so that may be enough in this context.

humitos: I opened https://github.com/readthedocs/website/pull/233

PR #200, opened 2023-07-26
Add platform URL for community/business on log in menus

### LINKED ISSUE #198: Login: mention site URL in the login dropdown
Should be a simple change, just adding a header to the list items for the current text and a subheader for the domain URL

agjohnson:
This adds a URL below the descriptive name to add a little more context
for the menu items.

Closes #198

----
:books: Documentation preview :books:: https://read-the-docs-website--200.com.readthedocs.build/

### DIFF
--- a/readthedocs_theme/templates/includes/topnav.html
+++ b/readthedocs_theme/templates/includes/topnav.html
@@ -47,11 +47,20 @@
   <a class="item" data-analytics="community-login" href="https://readthedocs.org/dashboard/">
     <i class="fad fa-people-group primary icon"></i>
     Read the Docs Community
+
+    <p class="ui mini grey text">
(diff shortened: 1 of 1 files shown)


ericholscher (approved):
This is 👍 

I'll probably steal a similar approach for the other places we're describing them (eg. pricing page, and adding content to this page describing each)

agjohnson:
I would probably still use the verbose names for most cases, but if we are pointing out to a link on each platform that would be an okay time to mention the domain.

I'm hoping to avoid getting too in the habit of calling it "Read the Docs for Business (readthedocs.com)", though we should help to draw the line for users a bit more. Ideally though, the domain name wouldn't be the important part that differentiates things for the user.

PR #201, opened 2023-07-27
Fix blog templates

agjohnson:
We did an original pass at the blog templates, but they needed clean up, fixes, and a better structure.

I jumped in here so we could at least start experimenting with putting all our website content in one place. This won't be super usable until we decide to do something with our newsletter integration though.

Some changes included here:

- Clean up the article display template
- Add category and tags to the display template
- Clean up the listing template
- Add category to the list template
- Use a macro for article cards on listing page, instead of repeating content
- Add category and tag listing templates, for linking to category pages and tag listing pages from the individual blog posts
- Fix the RSS/Atom feeds, only output one item per feed to make our newletter integration work
- Replace the default image with a generated SVG for now

Draft PR with some blog posts, to illustrate what this PR actually does: https://github.com/readthedocs/website/pull/202

----
:books: Documentation preview :books:: https://read-the-docs-website--201.com.readthedocs.build/

PR #202, opened 2023-07-27
Example blog posts

agjohnson:
For #201

----
:books: Documentation preview :books:: https://read-the-docs-website--202.com.readthedocs.build/


agjohnson:
This can be closed out after review, I don't think we necessarily want to bring in the posts here yet.

PR #203, opened 2023-07-31
Minimal port of the "Choosing between our platforms" content

ericholscher:
This just pulls the content in,
and makes it look decent.

I'll plan some additional content updates here, but wanted to try out a simple port first. 

Screenshot 2023-07-31 at 3 16 16 PM

Fixes https://github.com/readthedocs/website/issues/204


readthedocs_theme/templates/page.html
```diff
@@ -5,5 +5,9 @@
 {% block title %}{% if page.title %}{{ page.title|striptags }} - {% endif %}{{ SITENAME }}{%endblock%}
 
 {% block content %}
+<section>
+  <div class="ui padded container">
   {{ page.content }}
+  </div>
+</section>
```
ericholscher, line 12: @agjohnson there's still some weird spacing issues here at the end of the sections. Any thoughts? I just did a quick set of classes here to make the default display look pretty good, but it's still got some odd spacing.

agjohnson: Ah yes, so I hit this with the blog changes. This is the output from docutils and reST being slightly weird. Docutils will output a list of `div.section` elements, which is causing trouble with the default styles, which are assuming a more standard structure like:

```
<h1>...</h1>
<p>...</p>
<h2>...</h2>
<p>...</p>
```

I had updated a quick rule here specifically just for the blog pages:

https://github.com/readthedocs/website/blob/main/src/sui/themes/rtd-site/globals/site.overrides#L32

We can make these classes more of defaults, so no additional classes are required to active the styles when authoring reST content. But to make the pages look nice, we'll need some extra hacks to the content to output some special blocks. This additional control is why the more stylized pages just use HTML, it's easier than hacking up reST.

ericholscher: That makes sense. I'm 👍 on making it easy to draft default RST styles and make it look good, if it isn't hard.

ericholscher:
I've added the TOS & Privacy Policy in this PR as well, since that's content we've been wanting to port, and doesn't need any fancy display logic.

readthedocs_theme/templates/page.html
agjohnson: I looked at the styles more, it probably isn't a good fit to apply these styles across all pages, but we can still just use a separate template for this type of content. I put up an example here: 824456f

This uses a separate template that adds the class explicitly for docutils content, similar to what you did. But it adds the missing class for `.ui.docutils.container`, which styles the docutils output in a nicer way, and I also used `.ui.text.container`, which provides are larger font and narrower display. It makes the content look a bit more intentional when it's not as styled as our other pages.

ericholscher:
@agjohnson Thanks for that! I've updated the content of all the pages to properly resolve links, and updated the footer, and I think this PR should be ready go to!

humitos:
The rendered version is at https://read-the-docs-website--203.com.readthedocs.build/choosing-a-platform/ and it looks 👍🏼

PR #205, opened 2023-07-31
Update some configuration options to match prod

ericholscher:
Mainly:

* Remove RTD specific path logic that isn't needed
* Set proper SITEURL

----
:books: Documentation preview :books:: https://read-the-docs-website--205.com.readthedocs.build/


agjohnson (approved):

ericholscher:
@agjohnson I don't think we're actually using the publishconf anywhere, probably need to update the RTD command that is running.

PR #206, opened 2023-07-31
Rename About page to Features

ericholscher:
According to ChatGPT:

> The "About" page on a website serves as a space where the website owner or organization provides information about themselves, their purpose, and their mission. It is typically one of the essential pages on a website as it allows visitors to learn more about the website's identity, history, and values.

We want to talk about our product features, not about ourselves :)

I'd like to move the page to `/features/` as well, if we're 👍 on this change. 

This also rejiggers our footer to make it a bit cleaner:

Screenshot 2023-07-31 at 3 38 49 PM


agjohnson:
Originally, this page started off as "Product", but it wasn't clear enough what the user was supposed to do with the page. That name makes more sense when there are multiple products, and the "Products" link is some sort of dropdown. But a singular "Product" has a very different meaning and it doesn't really work in the same way.

"About" isn't a great fit though, but it seems better than "Product"

agjohnson:
Ah, yes, "Features" is a good upgrade :+1: I'd update the page name as well probably

ericholscher:
Cool, if you're 👍 I'll rename it and add redirects, and ship it 🎉

readthedocs_theme/templates/includes/footer.html
```diff
@@ -26,34 +26,35 @@ <h4 class="ui sub header">Stay updated</h4>
...
           </div>
         </div>
 
         <div class="column">
           <div class="ui vertical inverted text menu">
             {% block menu_services %}
-              <h4 class="ui sub header">Services</h4>
-              <a class="item" href="/pricing/">Pricing</a>
+              <h4 class="ui sub header">Content</h4>
               <a class="item" href="/about/">Features</a>
```
agjohnson, line 39: We can probably rename this page too, it doesn't need to stay "/about/", and I'm not sure we have any links to it directly either.

agjohnson (commented):
I'm :+1: on the page rename, but I actually think the layout of the footer links doesn't need to change as much. The existing structure seems okay, except for some of the links we've added for SEO.

readthedocs_theme/templates/includes/footer.html
ericholscher: Yea, that's the same page, I've pushed a PR renaming it.

readthedocs_theme/templates/includes/footer.html
```diff
@@ -26,34 +26,35 @@ <h4 class="ui sub header">Stay updated</h4>
...
             {% endblock menu_learn %}
           </div>
         </div>
 
         <div class="column">
           <div class="ui vertical inverted text menu">
             {% block menu_services %}
-              <h4 class="ui sub header">Services</h4>
-              <a class="item" href="/pricing/">Pricing</a>
+              <h4 class="ui sub header">Content</h4>
```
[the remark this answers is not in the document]
ericholscher: Yea, I'm not in love with this refactor, but I think it works reasonably well in that it makes it 4 lists of 4 items each. I'm not too worried either way if you feel strongly about any specific naming.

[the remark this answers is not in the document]
ericholscher: @agjohnson How about this:

* Removed Sphinx
* Renamed `Procuct`
* Moved Terms & Privacy back

Screenshot 2023-07-31 at 4 13 02 PM

I wanted all those links to be internal links, so I'll plan to migrate terms & privacy to the website after we figure out a pattern in #203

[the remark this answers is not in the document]
agjohnson: That looks good, it's more balanced vertically now too :+1:

readthedocs_theme/templates/includes/footer.html
agjohnson: Isn't this that PR? The link name is updated here, but it looks like the page is still `/about/`. I'm describing updating this link to `/features/` and moving `content/pages/about.html` to `content/pages/features.html`

agjohnson:
Also, the application footer will need updated too, but should be really straight forward. I've used Meld in the past, there are some differences due to Django URL resolving/etc.

content/pages/features.html
```diff
@@ -83,23 +83,48 @@ <h1 class="ui huge header">
   </div>
 </section>
 
-{% macro about_divide(text=None) %}
+<!-- First CTA -->
```
agjohnson, line 86: Template comments are used through other pages instead of HTML comments. We want these comments for internal notes, they aren't meant to be part of the output.

```suggestion
{# First CTA #}
```

content/pages/features.html
```diff
@@ -347,21 +372,22 @@ <h3 class="ui small header">
   </div>
 </section>
 
+<!-- Second CTA -->
```
agjohnson, line 375:
```suggestion
{# Second CTA #}
```

content/pages/features.html
```diff
@@ -347,21 +372,22 @@ <h3 class="ui small header">
...
           </p>
           <p>
             <button
                class="ui large inverted button"
-               onclick="jQuery('#about-modal').modal('show');"
+               onclick="jQuery('#features-modal').modal('show');"
                type="button">
               <i class="fad fa-play icon"></i>
-              Get started now
+              Get started free
```
agjohnson, line 390: This should at least be:

```suggestion
              Get started for free
```

agjohnson: Though I guess I normally try to avoid mentioning "free" for the negative value association/etc.

We don't really mention the trial anywhere major I think, so I lean towards something like this to obfuscate mentioning "free".

```suggestion
              Start your 1 month trial
```

Either one is fine though really.

content/pages/homepage.html
```diff
@@ -36,9 +36,10 @@ <h1 class="ui huge header">
                         data-analytics="signup-modal"
                          onclick="jQuery('#homepage-modal').modal('show');">
                         <i class="fad fa-play icon"></i>
-                        Get started
+                        Sign up now
                       </a>
-                      <a href="/about/" class="ui large stackable button">Learn more</a>
+                      <!-- Don't have multiple CTAs -->
+                      <!-- <a href="/features/" class="ui large stackable button">Learn more</a> -->
```
agjohnson, line 42: Or just remove this button too, that's also fine

```suggestion
                      {# Don't have multiple CTAs #}
                      {# <a href="/features/" class="ui large stackable button">Learn more</a> #}
```

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -60,11 +60,12 @@
       <code>https://readthedocs.com</code>
     </p>
   </a>
-  <div class="ui divider"></div>
-  <a class="item" href="{{ SITE_URL }}/about/">
+  <!-- Hide the learn more, we have a lot of Learn More links.. -->
+  <!-- <div class="ui divider"></div>
+  <a class="item" href="{{ SITE_URL }}/features/">
```
agjohnson, line 65: Also seems good to just remove this one, but if not:

```suggestion
  {# Hide the learn more, we have a lot of Learn More links.. #}
  {# <div class="ui divider"></div>
  <a class="item" href="{{ SITE_URL }}/features/">
```

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -60,11 +60,12 @@
...
     <i class="fad fa-circle-question grey icon"></i>
     Learn more
-  </a>
+  </a> -->
```
agjohnson, line 68:
```suggestion
  </a> #}
```

readthedocs_theme/templates/pages/features.html
agjohnson, line 1: Pretty sure this template isn't used at all and can just be removed. The features page here is using `template: page`

agjohnson (approved):
Just minor notes, everything looks in place for the rename now :+1:

readthedocs_theme/templates/includes/footer.html
agjohnson: Updates with the page source file rename and page edits look good :+1:

ericholscher:
Cleaned up based on feedback and added the redirect:
Screenshot 2023-08-01 at 9 33 38 AM

PR #208, opened 2023-08-01
Cleanup Header

ericholscher:
This does a few things:

* Removes the `Help` menu from the header, since those links are all in the footer, and link off site
* Adds the `Sign up` button to the header, so it's available on every page
* Makes Signup the most visual button, since that's what we most want
* Gives random pages proper titles where they were missing


agjohnson:
> Removes the Help menu from the header

This menu was kept here so that the application header and the website header were near identical. The plan has been to keep the header and the footer synced between the application and the website, so we have tried to keep the application in mind while structuring the website. I didn't yet port in the changes to reduce this item to an icon dropdown menu, but that would be a good change.

I would instead keep the help menu, but convert it to an icon button like in the application:

https://github.com/readthedocs/ext-theme/blob/main/readthedocsext/theme/templates/includes/header.html#L104-L106

This would preserve the structure and not make the headers competing. With this reduced to an icon menu, and the "Sign up" button a text link, there should be plenty of room in the header.

Eventually, this was going to be a menu in both the application and website, instead of just directly linking to the support page. I think it's helpful either way though. Other similar sites have various "Contact us" menus in the header, so this tracks:

image
> Adds the Sign up button to the header, so it's available on every page
> Makes Signup the most visual button, since that's what we most want

I would copy the element structure from the application, so that the headers are identical:

https://beta.readthedocs.org/accounts/login/

I prefer the text link for the sign in button, instead of competing buttons. There is also already a primary button for sign up above the fold on the front page, so I wouldn't use a primary button in the header too -- it pulls away attention on all the other pages.

> Adds the Learn more in the Signup dropdown back, linking to the choosing page

The mobile menu is a bit cluttered because of this I think:

image

At very least, we don't need the URLs on the second set of links, this is very redundant. Ideally, there shouldn't be competing links for log in and sign up though.

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -67,6 +70,27 @@
...
 
+{# Application links to sign up #}
+{% macro menu_sign_up() %}
+  <a class="item" data-analytics="community-signup" href="https://readthedocs.org/accounts/signup/">
+    <i class="fad fa-people-group primary icon"></i>
+    Read the Docs Community
+
+    <p class="ui mini grey text">
+      <code>https://readthedocs.org</code>
+    </p>
```
agjohnson, line 81: The sign up modal was meant to cover this particular issue, and the modal does a much better job than just a domain name at describing the platform's purpose and target audience. For that, I still think all sign up buttons should trigger the sign up modal instead linking directly to our websites.

This gives us a chance to describe things thoroughly, and maybe the user doesn't need to remember the domain name at all after being introduced to our platforms more thoroughly.

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -118,22 +147,34 @@
...
                   </div>
                 </div>
+              </div>
+
+              <div class="item">
+                <div class="ui fluid dropdown primary button" data-module="dropdown" data-action="select" data-display-type="block">
+                  <i class="fad fa-play icon"></i>
+                  <span class="text">Sign up</span>
+                  <div class="menu">
+                    {{ menu_sign_up() }}
```
agjohnson, line 175: I wouldn't use a dropdown specifically for sign up links. All sign up links should trigger the sign up modal so that we have a chance to describe the two platforms to the user.

content/pages/homepage.html
```diff
@@ -1,7 +1,6 @@
 <html>
   <head>
-    <!-- Intentionally left blank. Gives a truncated title in output -->
-    <title></title>
+    <meta name="title" content="Home" />
```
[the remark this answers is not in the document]
ericholscher: Trying out `Full featured documentation deployment platform`.

FWIW, `Home` is what the old site had, but I agree there's something better.

readthedocs_theme/templates/includes/topnav.html
ericholscher: Agreed. Any reason we don't have a login modal? Because presumably users know where they want to go once they're logging in?

ericholscher:
@agjohnson I pushed up the changes in the top nav:

Screenshot 2023-08-02 at 2 12 52 PM

I still don't love that the support page full takes you off the marketing website. I feel like it's really weird to have things in the header that are external links..

Overall I like this better (Signup goes to the modal, and having the login be more subtle.

agjohnson:
I do want to do something more with the support dropdown as we go, it should be a collection of support resources. And I would like to move resources that make more sense as marketing content to our website.

But, I think linking out for support resources is totally fine UX here. If a user is looking for support or help, they are not going to be interested in reading our marketing copy. This is a dead end of sorts for our funnel anyways, so serving the user's needs here works better for both of us really.

> I wasn't sure if you were asking for a Company/Contact us link in the header?

Ah no, just was poking around other sites and most similar sites have support/contact resources fairly prominent on their web pages.

readthedocs_theme/templates/includes/topnav.html
agjohnson: It's definitely an option, and was something I thought about trying originally, but wanted to make sure the modal pattern worked well first. I could see it annoying users too.

I landed at wanting to add some sort of platform detection (or naive platform detection like just storing a cookie when the user clicks signup/login for either community/commercial). This would make the login button either go directly to the correct site, or could instead only show a modal when there is no cookie.

Definitely some improvements here as we go, I'm mostly waiting to see what the usage looks like at the moment.

ericholscher:
@agjohnson Cool, should be ready for another review. I made it so:

* The `?` button does the existing dropdown on desktop
* There's still a `Help` menu on mobile

Which I think makes the most sense.

agjohnson:
The structure here is still using a dropdown menu item, so looks a bit janky, like it's a button in a button:

image

It looks like there are some extra elements and classes added

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -67,6 +70,19 @@
...
+{% macro menu_sign_up() %}
+  <a class="item" data-analytics="community-signup" href="https://readthedocs.org/accounts/signup/">
+    <i class="fad fa-people-group primary icon"></i>
+    Read the Docs Community
+  </a>
+  <a class="item" data-analytics="commercial-login" href="https://readthedocs.com/accounts/signup/">
+    <i class="fad fa-building secondary icon"></i>
+    Read the Docs for Business
+  </a>
+{% endmacro %}
```
agjohnson, line 83: Are we using this menu anywhere? Seems we can likely drop this entirely, in favor of the modal.

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -95,6 +111,11 @@
 
               <div class="divider"></div>
 
+              <div data-analytics="login-modal" class="header">Sign Up</div>
+              {{ menu_sign_up() }}
+
+              <div class="divider"></div>
```
agjohnson, line 117: We should still open the modal, even if it is on mobile/tablet view.

The modal was designed to work on all view ports:

image

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -118,30 +139,147 @@
...
+
+              <div class="item">
+                {# TODO: Add support page to marketing website #}
+                <div class="ui wide dropdown item" data-module="dropdown" data-action="select" data-display-type="block">
+                  <a aria-label="Help" class="ui circular basic icon button">
+                    <i class="fa-solid fa-message-question icon"></i>
+                  </a>
+                  <div class="menu">
+                    {{ menu_help() }}
+                  </div>
```
agjohnson, line 151: This is an `.item` inside a `.item` class, and is also wrapping the button in a `.ui.dropdown.item`. Both of these are causing the menu item to look like a button in a button.

image


Instead, the button should be a `.ui.dropdown.button`. There is an example of a icon dropdown menu in the docs:

https://fomantic-ui.com/modules/dropdown.html#pointing

I didn't test this, but something like:

```suggestion
                <div aria-label="Help" class="ui circular basic icon dropdown button" data-module="dropdown" data-action="select" data-display-type="block">
                  <i class="fa-solid fa-message-question icon"></i>
                  <div class="menu">
                    {{ menu_help() }}
                  </div>
                </div>
```

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -118,30 +139,147 @@
...
           </div>
         </div>
       {% endblock topnav_menu %}
 
     </div>
   </div>
 </div>
+
+
+{% block signup_modal %}
```
agjohnson, line 184: This modal will be defined in two places with this, again here:
https://github.com/readthedocs/website/blob/8095be9acdda237df82da32f39c23b292dc0bb37/content/pages/homepage.html#L42-L44

We should remove the content page source version, as we will already be defining this on every single page.

agjohnson:
A couple more tweaks for the mobile and tablet menu:

image

- The "Learn more" should maybe not have a divider above it at least, it's not clear what that link does. Grouping it with the other options might help? I think it's just an odd menu item though.
- The signup link in the mobile menu shouldn't be a button

ericholscher:
@agjohnson I fixed the stray `-`, but everything else feels very nitpicky? If you want to clean up the way you'd like it, that would be much quicker on my side, versus going back and forth on small UI changes. In particular the `Learn more` bit is the same between desktop & mobile, and splitting it out seems more annoying than having the divider? 🤷

agjohnson:
> @agjohnson I fixed the stray -, but everything else feels very nitpicky?

I get that this work is fiddly, but I'm not just picking things to change. The styles and patterns used in the navigation UI took time to establish and polish and I'm showing where we've regressed.

Being fluid around content changes on the site is *far* easier, especially copy and content, and I want be able to iterate quickly there.

But structural UI like this does need some patience and I probably would suggest planning and discussing changes beforehand in the future. This UI in particular had design constraints that we already worked through around content width, viewport width, and reuse of all of these patterns in the application UI. Planning up front is a good chance to discuss this, and figure out exactly what is changing and why.

> If you want to clean up

I'm always happy to jump in or answer questions. I definitely don't expect everyone to know how to fix or polish up all these UI issues. But I will tell folks to let me know when they need to tap out or need some help.

agjohnson:
With some of the changes to the menus:

[Screencast from 2023-08-08 17-29-47.webm](https://github.com/readthedocs/website/assets/1140183/32a12b2e-c8aa-432f-8e24-416c05818284)

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -135,7 +131,7 @@
 
               <div class="item">
                 {# TODO: Add support page to marketing website #}
-                <div class="ui wide circular basic icon dropdown button" data-module="dropdown" data-action="select" data-display-type="block">
+                <div class="ui circular basic icon wide pointing top right dropdown button" data-module="dropdown" data-action="select" data-display-type="block">
```
ericholscher, line 134: I had tried `pointing` previously, but didn't realize it required additional classes. I find it really hard to figure out what combination of classes work, and it also seems like the order matters, which I've also struggled with. But it's always obvious once there's an example, just very hard to actually get a working example for me.

agjohnson: Aye, ordering *can* be important for some of the rules, though most are okay in any order. There are some directional classes like `top right` that I think might be order specific. I do try keep classes in some order of ascending importance, but this is very subjective and not a firm rule.

Pointing menu directions aren't covered super well in the docs that I can find. The variant is covered here, and does mention the extra classes in the examples, but does not do a good job of calling out how to use them:
https://fomantic-ui.com/modules/dropdown.html#pointing

I feel like this is a gap in the docs and I probably know this from working with another similar element like popups or something.

PR #211, opened 2023-08-01
Refactor marketing copy on homepage

ericholscher:
This does a pretty sizable refactor to the messaging and what we're highlighting.
I think it's a good step forward,
but definitely not perfect by any means.

Any input is helpful.
The best way to review it likely with the PR preview. 

----
:books: Documentation preview :books:: https://read-the-docs-website--211.com.readthedocs.build/


content/pages/homepage.html
```diff
@@ -112,28 +111,33 @@ <h3 class="ui large monospace header">
...
+            {% call homepage.feature_detail_item(header="Supports any doc tool", icon="fad fa-terminal") %}
+              Integrated support for Sphinx, MkDocs, and Jupyter Book.
+              Flexible enough to work with any documentation tool.
             {% endcall %}
 
-            {% call homepage.feature_detail_item(header="Versioned docs", icon="fad fa-layer-group") %}
-              Match your product release cycle with multiple versions of your docs. Automatically build versions from any branch or tag.
+            {% call homepage.feature_detail_item(header="Integrated search", icon="fad fa-search") %}
+              Easily search all your docs via our dashboard,
+              and offer your users a more powerful search with our  search API.
```
agjohnson, line 121: I'd probably leave dashboard and API out of this description, that's more technical than we need to describe the feature. We describe searching across all projects, which is a good improvement here, we could describe more search features too instead:

```suggestion
              Easily search across all of your projects and
              control search results with fine-grained configuration.
```

content/pages/homepage.html
```diff
@@ -112,28 +111,33 @@ <h3 class="ui large monospace header">
...
             {% endcall %}
 
-            {% call homepage.feature_detail_item(header="Build constantly", icon="fad fa-gears") %}
-              Preview every pull request, and build with every commit. Keep your docs up to date, automatically.
+            {% call homepage.feature_detail_item(header="Pull request previews", icon="fad fa-gears") %}
```
agjohnson, line 124: Icon could use updated here

content/pages/homepage.html
```diff
@@ -112,28 +111,33 @@ <h3 class="ui large monospace header">
...
+              Preview each pull request with a unique URL,
+              allowing fast collaboration across your organization.
```
agjohnson, line 126: The user probably doesn't have questions about the preview URL, so this might bring up questions instead of clarify.

```suggestion
              Preview each pull request before releasing any changes,
              allowing for fast collaboration across your organization.
```

content/pages/homepage.html
```diff
@@ -112,28 +111,33 @@ <h3 class="ui large monospace header">
...
-              Help your users find what the need with integrated search across all your projects and versions.
+            {% call homepage.feature_detail_item(header="Versioned docs", icon="fad fa-layer-group") %}
+              Match your product release cycle with multiple versions of your docs. Automatically build versions from any branch or tag.
             {% endcall %}
 
-            {% call homepage.feature_detail_item(header="Private cloning", icon="fad fa-lock", paid=True) %}
-              Clone private repositories using a connected GitHub, GitLab, or Bitbucket account.
+            {% call homepage.feature_detail_item(header="Private repository support", icon="fad fa-lock", paid=True) %}
+              Clone private repositories using a connected GitHub account,
+              or any Git provider with an SSH key.
```
agjohnson, line 135: Should either mention all or none of the providers here I guess. This sounds like only GitHub is supported.

```suggestion
              Clone private repositories using a connected GitHub, GitLab, or Bitbucket account,
              or clone from any Git provider with an SSH key.
```

content/pages/homepage.html
```diff
@@ -112,28 +111,33 @@ <h3 class="ui large monospace header">
...
             {% endcall %}
 
             {% call homepage.feature_detail_item(header="Authentication", icon="fad fa-fingerprint", paid=True) %}
-              Control who has access to your documentation with integrated authentation.
+              Flexible permission system built in.
```
agjohnson, line 139: I think the original was a bit more clear here. "Permission system" doesn't imply documentation authentication to me. I'd keep the second sentence about SSO though. We do mention that somewhere else already I thought but another doesn't hurt.

content/pages/homepage.html
```diff
@@ -154,97 +158,200 @@ <h3 class="ui large monospace header">
...
+                  # nodejs: "16"
+    
+              # Build documentation in the docs/ directory with Sphinx
+              sphinx:
+                 configuration: docs/conf.py
+    
+              # Dependencies required to build your docs
+              python:
+                 install:
+                 - requirements: docs/requirements.txt
```
agjohnson, line 196: Great call, this is a great block for the front page. This would be another place to make these working examples instead, right now this is vaguely correct but won't work if the user tries to do anything with it.

agjohnson (commented):
I didn't poke through all of these changes visually yet. I noted a few areas where the icon will not match copy, and other areas that are easy to miss are narrow and tablet sized viewports. The narrow viewport can be an issue for wide columns and a few areas where we purposely kept copy short to avoid wrapping issues.

content/pages/homepage.html
```diff
@@ -22,13 +24,11 @@
                 {% block homepage_callout %}
                   <div class="ui large left aligned basic vertical segment">
                     <h1 class="ui huge header">
-                      Write documentation,<br />
-                      we'll do the rest
+                      You write the docs,<br />
+                      we handle the rest.
```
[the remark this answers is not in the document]
ericholscher: Yea, I definitely was struggling with the "what does your product actually do?" question. It's hard to sell benefits to users while also explaining what your product does, but I agree this header doesn't really pass the test.

content/pages/homepage.html
```diff
@@ -22,13 +24,11 @@
...
                     </h1>
                     <p>
-                      Make documentation part of your development workflow with
-                      simplified processes that automate building, versioning,
-                      and hosting of your docs.
+                      Automate your documentation workflow with our full-featured documentation deployment solution.
```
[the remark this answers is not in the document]
ericholscher: Yea, I know we've using `Building and hosting` in the past, which is what deployment captures. I think `workflow` is probably another term to use?

content/pages/homepage.html
ericholscher: I was trying to convey both the dashboard & in-doc search. It's really hard to talk about the in-doc search, comparing it against their doc tools search, and API was the best way I could think of to talk about this.

Perhaps it's getting too deep though :)

content/pages/homepage.html
ericholscher: Agreed -- would be neat to have a "deploy this example" button, and they just need to login with GH, and it magically deploys docs. We already have the example projects, so it wouldn't be hard to configure this.

content/pages/homepage.html
```diff
@@ -64,37 +64,36 @@ <h1 class="ui huge header">
...
-            Automatically release your documentation with every commit,
-            and preview documentation changes with every pull request.
+            Preview changes on every commit to your pull requests.
+            Release documentation to your users on each merge.
           </p>
         {% endcall %}
 
+        {% call marketing.highlight_item(
+          icon="olive fa-terminal",
+          header="Ideal developer experience",
```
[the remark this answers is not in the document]
ericholscher: Yea, I'm trying to clarify the benefit over the noun of what it is. Not 100% in love with mine, but "developer-driven documentation" doesn't really tell me the benefit (is it good, presumably?)

content/pages/homepage.html
```diff
@@ -64,37 +64,36 @@ <h1 class="ui huge header">
...
+          color="black") %}
+        <p class="small" >
+          Write documentation without changing your workflow or your tools
+          using <strong>docs as code</strong>.
```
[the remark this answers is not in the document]
ericholscher: Yea, we have #194 which I want to finish up.

content/pages/homepage.html
ericholscher: Yea, it's a core feature, moved back the original copy.

ericholscher (commented):
@agjohnson I pushed up the obvious changes and merged most of your suggestions. ~3 comments left that I wasn't sure about moving forward on.

content/pages/homepage.html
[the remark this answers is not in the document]
ericholscher: I pushed a change with:

```
Documentation Simplified
Build, host, and share documentation in seconds
```

I think this captures the product and value.

content/pages/homepage.html
ericholscher: For what it's worth, it's basically the exact config we have in our example Sphinx project:

https://github.com/readthedocs-examples/example-sphinx-basic/

I thought our onboarding prompts users to install this right, so it should mostly work?

agjohnson: A one click deploy option would be really neat. I don't know what we have configured for that sort of workflow currently, though know the tutorial at least promotes the template repository at least.

I was suggesting this all as a next step, so nothing block here of course. These would be good options to explore for later though

content/pages/homepage.html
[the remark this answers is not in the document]
agjohnson: Aye, true.

Maybe I'm thinking something closer to "Documentation developers love"? Though neither answers exactly why developers are important in the equation.

content/pages/homepage.html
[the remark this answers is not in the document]
agjohnson: Yeah, workflow seems like a good one to keep mentioning. "Deployment" does still work well, probably better than "building", just need to work in a hosting mention somehow.

content/pages/homepage.html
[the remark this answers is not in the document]
agjohnson: Weird, the review got disconnected here. I'll continue at the new review location, but this is on a good track

ericholscher:
Gonna also merge this one, since I want this to go live for our .org redirect.

agjohnson:
I'll have some notes here on the top CTA that I think we should address.

PR #214, opened 2023-08-08
Some fixes for the front page header changes

agjohnson:
Some leftovers from #211. Fixes capitalization to match our other
headers and puts the focus not on time but on our service.

I've had this feedback on other similar CTAs. Focusing on
minutes/seconds is saying our value is that we save time. That's not
wrong, but we don't exactly aim to save time, but rather our value is
that we manage all of the hurdles with infrastructure when building and
hosting authed/searchable/etc docs.

Both versions here could also use more content, as it looks less good, and quite sparse, with only two lines of text:

image

----
:books: Documentation preview :books:: https://read-the-docs-website--214.com.readthedocs.build/

### DIFF
--- a/content/pages/homepage.html
+++ b/content/pages/homepage.html
@@ -42,10 +42,10 @@
                 {% block homepage_callout %}
                   <div class="ui large left aligned basic vertical segment">
                     <h1 class="ui huge header">
-                      Documentation Simplified
+                      Documentation simplified
                     </h1>
                     <p>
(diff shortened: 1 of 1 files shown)


ericholscher (approved):

ericholscher:
Yea, I think the timing aspect of the CTA is around how easy it is to get started, to try to entice folks. I agree it shouldn't be in the subhead though :)

PR #220, opened 2023-08-10
Hack to use 404 error page in RTD

### LINKED ISSUE #111: Configure error pages with RTD

agjohnson:
I opened https://github.com/readthedocs/readthedocs.org/issues/10622 to
consider making this configurable. The other error status codes are not
supported by RTD yet.

- Closes #111
- Refs https://github.com/readthedocs/readthedocs.org/issues/10622

----
:books: Documentation preview :books:: https://read-the-docs-website--220.com.readthedocs.build/

### DIFF
--- a/content/pages/errors/404.html
+++ b/content/pages/errors/404.html
@@ -3,7 +3,9 @@
     <title>Not found</title>
     <meta name="subtitle" content="The resource you requested is not recognized" />
     <meta name="template" content="error" />
-    <meta name="slug" content="errors/404" />
+    {# The path for 404 pages is not flexible and needs to be top-level #}
+    {# <meta name="slug" content="errors/404" /> #}
+    <meta name="slug" content="404" />
     <meta name="status" content="hidden" />
     <meta name="icon" content="fa-radar" />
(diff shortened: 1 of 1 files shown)


agjohnson:
Example 404 page at: https://read-the-docs-website--220.com.readthedocs.build/foo

I'd like to update that one a little bit, it was meant more as an application error page. But it works at least and you don't get the RTD search page.

ericholscher (approved):

ericholscher:
I see we're actually outputting these files, which we probably don't want to do? eg. https://about.readthedocs.com/errors/401/

agjohnson:
> I see we're actually outputting these files, which we probably don't want to do?

This was indeed intentional, but if there is a good reason not to we don't have to. We don't really have another option though, yeah. Maybe this is all a good discussion for an improved version of the error response overrides on RTD.

Ideally, I'd love to configure RTD to know about all of these templates, configure them at a nested path, and not serve them through regular requests like `/errors/404/`. I'd also like to configure other 4xx and 5xx responses too.

PR #228, opened 2023-09-08
Improve buttons in sign up modal

humitos:
I found that the signup modal doesn't link to the "Choosing a platform" page, which I think it's important in that context. The modal shows small differences between the two sites, but I felt there was some context missing there. I added a link to this page so users can dig deeper if they want to.

Besides, I changed the copy from "Learn more" to "Explore all the features", that communicates exactly what that button is for.

Finally, I added icons to these buttons because they make everything to look better.

----
:books: Documentation preview :books:: https://read-the-docs-website--228.com.readthedocs.build/

This is how it looks with these changes:

Screenshot_2023-09-08_12-55-32

### DIFF
--- a/readthedocs_theme/templates/includes/topnav.html
+++ b/readthedocs_theme/templates/includes/topnav.html
@@ -266,7 +266,8 @@
     </div>
 
     <div class="actions">
-      <a href="{{ SITE_URL }}/features/" class="ui cancel button">Learn more</a>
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
This looks great :+1: 

I think this whole UI *may* be duplicated somewhere in the templates or page content. It would be worth double checking, but I recall consolidating these all at some point too.

humitos:
> I think this whole UI _may_ be duplicated somewhere in the templates or page content. It would be worth double checking, but I recall consolidating these all at some point too.

I didn't find this content duplicated, so I think we are 👍🏼

PR #236, opened 2023-09-18
GitHub: add "No usage limits"

humitos:
----
:books: Documentation preview :books:: https://read-the-docs-website--236.com.readthedocs.build/


agjohnson (approved):

stsewd:
If we are thinking about having limits, we should maybe wait till we decide on those before making this change?

humitos:
We don't limits right know, so I think this matches what we offer. We can change it once we have those limits deployed.

PR #244, opened 2024-01-18
Show Business as "Ad free" so that checks are always "good"

ericholscher:
This makes it more obvious that Community had ads, and that's a "bad thing"

----
📚 Documentation preview 📚: https://read-the-docs-website--244.com.readthedocs.build/

### DIFF
--- a/content/pages/pricing.html
+++ b/content/pages/pricing.html
@@ -127,7 +127,7 @@
                   {{ pricing.plan_feature("Public projects", "Host public documentation", "fa-globe") }}
                   {{ pricing.plan_feature("Public repositories", "Clone public repositories", "fa-globe") }}
                   {{ pricing.plan_feature("Custom domain", "Use your own domain", "fa-at") }}
-                  {{ pricing.plan_feature("Ad supported", "Hosted documentation includes ads", "fa-ad") }}
+                  {{ pricing.plan_feature("Ad supported", "Hosted documentation with ads", "fa-ad") }}
                 </div>
               </div>
             </div>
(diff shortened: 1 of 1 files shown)


content/pages/pricing.html
```diff
@@ -219,8 +219,8 @@ <h3 class="ui small header">
     {% call pricing.feature_table() %}
       {{ pricing.feature_row("Team management", "Control organization access with team permissions", "fa-users", on_community=False) }}
       {{ pricing.feature_row("Email support", "Get help with your projects and documentation", "fa-message-question", "Best effort", "2 business days", "2 business days", ("1 business day", "olive")) }}
+      {{ pricing.feature_row("Ad free", "Documentation without ads", "fa-ad", on_community=False) }}
```
agjohnson, line 222: I'd also find a more specific phrase to use here. I think the reason this feels less clear is that this changing this to "with"/"without" removes the verb from the phrase.

agjohnson (commented):
This makes sense, I get the psychology on the change. Also, we can alter the icons for this row if we want. It would be clearer without the double negative.

content/pages/pricing.html
```diff
@@ -127,7 +127,7 @@ <h2 class="ui small monospace center aligned header">
                   {{ pricing.plan_feature("Public projects", "Host public documentation", "fa-globe") }}
                   {{ pricing.plan_feature("Public repositories", "Clone public repositories", "fa-globe") }}
                   {{ pricing.plan_feature("Custom domain", "Use your own domain", "fa-at") }}
-                  {{ pricing.plan_feature("Ad supported", "Hosted documentation includes ads", "fa-ad") }}
+                  {{ pricing.plan_feature("Ad supported", "Hosted documentation with ads", "fa-ad") }}
```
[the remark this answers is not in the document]
ericholscher:
```suggestion
                  {{ pricing.plan_feature("Ad supported", "Hosted documentation includes ads", "fa-ad") }}
```

content/pages/pricing.html
ericholscher:
```suggestion
      {{ pricing.feature_row("Ad free", "Documentation doesn't include ads", "fa-ad", on_community=False) }}
```

Perhaps just inverting the original?

agjohnson: The sounds clearer. We've also used "documentation is ad-free" on our marketing or docs pages too.

PR #243, opened 2024-01-17
Mention free trial on pricing page

ericholscher:
Also add a business callout on the Community tab, so we can link folks to it directly.

Fixes https://github.com/readthedocs/website/issues/185

## Screenshots

Screenshot 2024-01-17 at 3 54 04 PM

Screenshot 2024-01-17 at 3 54 07 PM

----
📚 Documentation preview 📚: https://read-the-docs-website--243.com.readthedocs.build/

PR #245, opened 2024-01-18
Fix change to commercial callout

agjohnson:
In #243 this was changed, but it now implies projects have to be private on commercial.
This reverts the change, and reuses terms we are already using on the
page: "private and commercial documentation/projects".

https://github.com/readthedocs/website/pull/243/files#r1457943722

----
📚 Documentation preview 📚: https://read-the-docs-website--245.com.readthedocs.build/


ericholscher:
Thanks for the post-merge review 👍

PR #252, opened 2024-02-01
Add title to blog index page and image to newsletter

ericholscher:
----
📚 Documentation preview 📚: https://read-the-docs-website--252.com.readthedocs.build/


ericholscher:
Gonna merge this so it goes out with the newsletter, but post-merge review should be fine here.

readthedocs_theme/templates/article.html
```diff
@@ -53,7 +53,7 @@
 
       <div class="ui divider"></div>
 
-      <img class="ui big rounded centered image" src="{{ article.image or '/images/posts/default.svg' }}" alt="{{ article.image_credit }}">
+      <img class="ui big rounded centered image" src="{{ article.image or '/images/posts/default.svg' }}" alt="{{ article.image }}">
```
agjohnson, line 56: Looks like the alt was replaced with the image URL. I think this was correct before, no?

agjohnson: Yeah, the issue comes from using HTML in the image_credit, which isn't valid in an HTML attribute. I would use `striptags` or escape the HTML instead of replacing with the URL

PR #253, opened 2024-02-02
Fix article image alt attribute

agjohnson:
PR #252 dropped the alt text to support HTML in the image credits, but a
more accessible fix here is just strip the HTML from the image credits
when using the credits in the attribute.

----
📚 Documentation preview 📚: https://read-the-docs-website--253.com.readthedocs.build/

### DIFF
--- a/readthedocs_theme/templates/article.html
+++ b/readthedocs_theme/templates/article.html
@@ -53,7 +53,7 @@
 
       <div class="ui divider"></div>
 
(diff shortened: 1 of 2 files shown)


ericholscher (approved):

ericholscher:
This isn't really explaining the image, it's just a credit. I don't think this is really useful for an alt tag, but 🤷

agjohnson:
Yeah sure, but it's better than a URL I feel.

PR #254, opened 2024-02-02
Remove RSS feed

ericholscher:
----
📚 Documentation preview 📚: https://read-the-docs-website--254.com.readthedocs.build/

### DIFF
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -1,10 +1,18 @@
 """Read the Docs website Pelican configuration."""
 
+import os
 
 AUTHOR = 'Read the Docs, Inc'
 SITENAME = 'Read the Docs'
-SITEURL = ''
 RELATIVE_URLS = True
+
+if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
+    # Make all URLs "domainless" in the RTD PR preview
+    # https://docs.readthedocs.io/en/stable/reference/environment-variables.html
+    SITEURL = ""
+else:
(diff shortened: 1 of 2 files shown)


agjohnson:
I mentioned in Front we should fix the site URL, we sort of broke that at some point by making it a relative URL. This affects feed URLs, but also other links in HTML. Ethicalads uses this:

https://github.com/readthedocs/ethicalads.io/blob/main/publishconf.py#L13-L19

We should use something similar in pelicanconf.py, or go back to using publishconf.py.

ericholscher:
Yea, that looks like a better solution 💯

agjohnson (approved):

pelicanconf.py
```diff
@@ -1,11 +1,19 @@
...
-SITEURL = ''
 RELATIVE_URLS = True
 
+if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
+    # Make all URLs "domainless" in the RTD PR preview
+    # https://docs.readthedocs.io/en/stable/reference/environment-variables.html
+    SITEURL = ""
+else:
+    # This setting is needed to make the RSS/Atom feeds generate correctly
+    SITEURL = "https://https://about.readthedocs.com"
```
humitos, line 15: Is the double "https://” correct here?

agjohnson: It's for extra security.

agjohnson: I put up a fix at #257 :smile:

PR #261, opened 2024-02-15
Use filename as slug source

stsewd:
By default, the title of the article is used as the slug, if the title is changed, the slug will be updated, and the URL of the article will be changed, leading to a 404.

Our previous blog used the filename as the slug, I like that.

If we agree, I can set slug explicitly for the articles that are already published or maybe create a redirect.

----
📚 Documentation preview 📚: https://read-the-docs-website--261.com.readthedocs.build/

### DIFF
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -68,6 +68,7 @@ ARTICLE_PATHS = ['posts']
 STATIC_PATHS = ['images']
 
 # URL settings
+SLUGIFY_SOURCE = "basename"
 PAGE_URL = '{slug}/'
 PAGE_SAVE_AS = '{slug}/index.html'
 ARTICLE_URL = 'blog/{date:%Y}/{date:%m}/{slug}/'


agjohnson (approved):
Good catch! I agree with using the basename, we've gone back to update titles in the past.

stsewd:
Got the diff from all the filenames before and after the change

```diff
$ diff a.txt b.txt
7,8c7,8
< output/blog/2024/02/read-the-docs-newsletter-february-2024
< output/blog/2024/02/read-the-docs-newsletter-february-2024/index.html
---
> output/blog/2024/02/newsletter-february-2024
> output/blog/2024/02/newsletter-february-2024/index.html
32a33,34
> output/drafts/2022-war-in-ukraine.html
> output/drafts/newsletter-march-2022.html
37,38d38
< output/drafts/read-the-docs-newsletter-march-2022.html
< output/drafts/war-in-ukraine-and-what-it-means-for-read-the-docs.html
```

Only the newsletter needs a redirect, I'll create one.

stsewd:
Also, I'm still deleting branches manually, can we enable the "delete heads branches after merge" option in the repo? :)

agjohnson:
Yeah, feel free to enable. I don't think we have a reason not to, just never set it up.

stsewd:
@agjohnson I don't have permissions on this project.

agjohnson:
Oops, try now. I added core team with admin permissions.

stsewd:
All set now :+1:

PR #263, opened 2024-02-15
Don't use relative URLs when deploying

stsewd:
See https://docs.getpelican.com/en/latest/settings.html?highlight=RELATIVE_URLS#url-settings.

This should be used when developing locally only.

Closes https://github.com/readthedocs/website/issues/226

----
📚 Documentation preview 📚: https://read-the-docs-website--263.com.readthedocs.build/

### DIFF
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -4,15 +4,15 @@ import os
 
 AUTHOR = 'Read the Docs, Inc'
 SITENAME = 'Read the Docs'
-RELATIVE_URLS = True
 
-if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
-    # Make all URLs "domainless" in the RTD PR preview
-    # https://docs.readthedocs.io/en/stable/reference/environment-variables.html
-    SITEURL = ""
-else:
-    # This setting is needed to make the RSS/Atom feeds generate correctly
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
I can't say if this is going to have any unwanted side effects, but I like your solution much better. If it isn't already, we should document this configuration, as I'm sure it's common.

stsewd:
> If it isn't already, we should document this configuration, as I'm sure it's common.

Do you mean in this repo? Or on RTD docs?

agjohnson:
Oh sorry, yeah in our user docs. Unfortunately, we only expose this in our docs in the build commands page, so this configuration is a bit out of scope for that page:

https://docs.readthedocs.io/en/stable/build-customization.html#pelican

I think we were trying to use another repository to show some examples though, maybe there instead?

@humitos where were we documenting examples for build tools?

humitos:
> @humitos where were we documenting examples for build tools?

Not sure, but you may refer to https://github.com/readthedocs-examples/ organization. It currently has Sphinx, MkDocs and Jupyter Book examples. A contributor added Antora at (https://github.com/man-chi/example-antora-basic/) that we should migrate into our organization somehow to gain control.

That organization is a good place to create a Pelican example if we want to document this somewhere.

PR #276, opened 2024-03-16
Fix "because we ❤️ documentation" FA icon

janbrasna:
Probably overlooked or not updated to the current markup.

Fixes `fas heart` icon inside a sentence.
\+ removes from ARIA roles
\+ updates branding link

Screen Shot 2024-03-16 at 2 14 29

----
📚 Documentation preview 📚: https://readthedocs-about--276.org.readthedocs.build/


agjohnson (approved):
Great catch, thanks for the fix! These updates all look good to me

PR #283, opened 2024-05-14
Remove "external link" icon on blog, and put it at top of resources

ericholscher:
----
📚 Documentation preview 📚: https://readthedocs-about--283.org.readthedocs.build/


agjohnson:
The change looks okay technically, but I think the user is much more likely to click through to documentation, support, or our tutorial. This is the reason they are at the top of the list currently, which still feels correct.

PR #295, opened 2024-06-16
Pricing: highlight advanced plan

### LINKED ISSUE #288: Pricing: highlight "Advanced plan"
We want to add some kind of distinction for the "Advanced plan". Show it with a different color or similar. Besides, add a "Recommended" or "Most valued" or similar...

This is the page we are talking about https://about.readthedocs.com/pricing/

humitos:
Initial proposal to highlight advanced plan.

Screenshot_2024-06-16_23-17-38

Closes #288

----
📚 Documentation preview 📚: https://readthedocs-about--295.org.readthedocs.build/

### DIFF
--- a/content/partials/pricing.html
+++ b/content/partials/pricing.html
@@ -8,7 +8,7 @@
 {#
   Main plan block for commercial/community plan cards
 #}
-{% macro plan(name, price, color="teal") %}
+{% macro plan(name, price, color="teal", highlight=False) %}
   <div class="{% if not price %} computer only {%- endif %} column">
     <div class="ui {%- if price %} raised {{ color }} {%- else %} basic {%- endif %} segment">
 
@@ -32,6 +32,12 @@
       {% endif %}
 
       {{ caller() }}
+
+        {% if highlight %}
+        <div class="ui teal bottom attached label">
+          Most popular
+        </div>
+        {% endif %}
 
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


humitos:
I still think the ribbon version is a lot better, but I wasn't able to deal with the vertical spacing/alignment as @agjohnson anticipated 😅 

Screenshot_2024-06-17_09-45-51

Is there a simple way to solve this?

agjohnson:
Yup, that is the issue I mentioned you would run into trying to use the ribbon style. There isn't a way around this, as it takes up vertical space.

ericholscher:
Could put the ribbon above the pricing box, and just move them all down. But yea, not 100% sure the best option.

agjohnson:
The ribbon style only really works if all the cells in the row have a similar ribbon, which makes this note not stand out. I had assigned this issue to myself to play with some better options here, but think we can start with something basic instead.

content/partials/pricing.html
```diff
@@ -33,6 +33,12 @@ <h2 class="ui small monospace center aligned header">
 
       {{ caller() }}
 
+        {% if highlight %}
+        <div class="ui teal bottom attached label">
+          Most popular
```
agjohnson, line 38: I'd make this at least more wordy.

```suggestion
          Our most popular plan
```

You might also add an icon to improve the visibility of this though.

ericholscher:
Yea, I think the original concept is a reasonable first step, or a border or similar. Lots of options, just need to ship a v1 and then iterate when we have other ideas 👍

humitos:
Screenshot_2024-06-18_10-35-21

agjohnson (approved):

PR #296, opened 2024-06-18
Upgrade NPM dependencies

humitos:
Attempt to upgrade all the NPM dependencies and NodeJS as well. There are some issues with the fonts that I'm not sure why they are happening. I guess it's because of some static/public paths that I'm missing when migrating from Webpack 4 to 5.


humitos:
It would be good to move forward with this work and try to keep the dependencies updated as we do in other projects. @agjohnson do you know what I'm missing here? I suppose it's something pretty basic that I'm not seeing.

agjohnson:
Not sure what would be the case here without spending time here. This feels lowish priority though, so probably won't make it on my list for a while.

PR #298, opened 2024-06-18
Add logos of our users to our homepage

ericholscher:
Fixes https://github.com/readthedocs/website/issues/285

Not 100% sure the best way to make these line up. Is there a standard way? 

----
📚 Documentation preview 📚: https://readthedocs-about--298.org.readthedocs.build/


content/includes/business-logos.html
```diff
@@ -0,0 +1,19 @@
+<section id="business-logo">
+    <div class="ui very padded container">
+        <div class="ui four column grid">
```
agjohnson, line 3: This should be `.ui.stackable.grid`, or it will try to display inline on mobile viewports:

image

content/includes/business-logos.html
```diff
@@ -0,0 +1,19 @@
...
+            <div class="column">
```
agjohnson, line 4: The logos appear out of alignment in part because they are top aligned inside the column. You can set the column to middle alignment with:

```suggestion
            <div class="middle aligned column">
```

content/includes/business-logos.html
```diff
@@ -0,0 +1,19 @@
...
+                <img class="ui small image" src="{{ SITE_URL }}/images/amd-logo.svg" alt="AMD Logo">
```
agjohnson, line 5: This doesn't need capitalization here and instances below:

```suggestion
                <img class="ui small image" src="{{ SITE_URL }}/images/amd-logo.svg" alt="AMD logo">
```

content/includes/business-logos.html
```diff
@@ -0,0 +1,19 @@
...
+            </div>
+            <div class="column">
+                <img class="ui small image" src="{{ SITE_URL }}/images/canonical-logo.svg" alt="Canonical Logo">
+            </div>
+            <div class="column">
+                {# This is medium because the logo is small otherwise? :shrug: #}
```
agjohnson, line 11: This is mainly because the images are all of varying dimensions. It's best to edit the images down to a consistent dimension for this.

Also note that `.ui.small.image` scales the image to the _width_ dimension, which is why it looks especially small in comparison

content/includes/business-logos.html
```diff
@@ -0,0 +1,19 @@
+<section id="business-logo">
```
agjohnson, line 1: This should probably have a heading too

ericholscher:
With these changes, I think it looks shippable:

Screenshot 2024-06-19 at 1 19 29 PM

We can definitely iterate on it, but I think it definitely improves the landing page already.

ericholscher:
@agjohnson Would love another look here, would be good to get it shipped!

content/pages/homepage.html
```diff
@@ -66,11 +66,11 @@ <h1 class="ui huge header">
         {% call marketing.highlight_item(
             icon="white fa-user-robot",
             icon_style="--fa-secondary-color: skyblue; --fa-secondary-opacity: 1.0",
-            header="Easy previews and deploys",
+            header="Automatic deploy previews",
             color="grey") %}
           <p>
-            Preview changes on every commit to your pull requests.
-            Release documentation to your users on each merge.
+            Preview your entire documentation on every pull request.
```
agjohnson, line 72:
```suggestion
            Preview your documentation on every pull request.
```

content/pages/homepage.html
```diff
@@ -90,13 +90,17 @@ <h1 class="ui huge header">
             header="Work privately or publicly",
             color="violet") %}
           <p>
-            Easily share within your team or with the whole world.
-            Manage permissions yourself or automatically with GitHub.
+            Share content with just your team, or with the whole world.
+            Sync permissions from GitHub, or use custom permissions in our dashboard.
```
agjohnson, line 94: We should avoid mentioning the dashboard here probably, as permissions extend to docs too.

```suggestion
            Sync permissions from GitHub, or manage custom permissions for your projects.
```

agjohnson (commented):
Looking close but this still needs centered on mobile:

image

Likely this is using `.center.aligned.column` on the surrounding column with `.ui.inline.image` on the logo images.

ericholscher:
Got it working:
Screenshot 2024-07-08 at 2 06 37 PM

content/includes/business-logos.html
```diff
@@ -0,0 +1,21 @@
...
+    <div class="ui very padded container">
+        <h3 class="ui huge center aligned column header">
```
agjohnson, line 3: This isn't needed.

```suggestion
        <h3 class="ui huge center aligned header">
```

content/includes/business-logos.html
```diff
@@ -0,0 +1,21 @@
...
+            Trusted by businesses worldwide since 2010
+        </h3>
+        <div class="ui padded stackable four column grid">
+            <div class="center aligned column">
```
agjohnson, line 7: You still want `.middle.aligned` class here, as the images were not resized to the same dimensions. The vertical alignment is off without this:

image

```suggestion
            <div class="middle aligned center aligned column">
```

agjohnson (commented):
Close, but the removed classes were still needed for vertical alignment. All of the columns need the same classes.

agjohnson (approved):
Alignment looks good now :+1:

ericholscher:
🎉

PR #300, opened 2024-06-21
Small updates to choosing platform

humitos:
- Only mention Git as supported VCS
- Add buttons to sign up instead of small links (this follow the same pattern than on the pricing page)

----
📚 Documentation preview 📚: https://readthedocs-about--300.org.readthedocs.build/choosing-a-platform/

### DIFF
--- a/content/pages/choosing-a-platform.rst
+++ b/content/pages/choosing-a-platform.rst
@@ -25,14 +25,21 @@ services, for projects of all sizes.
 Important points:
 
 * Open source project hosting is always free
-* All documentation sites include advertising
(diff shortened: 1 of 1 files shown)


ericholscher (approved):
We should probably remove this page and just point folks at the pricing page, tbh. I think this made more sense when there wasn't a feature checklist, but now we have a great one.

humitos:
Yeah, this page is not great. I think it would be a good idea to merge with the pricing one. IMO, the pricing page should have a similar description explaining this relevant points probably.

PR #303, opened 2024-06-26
Fix casing on enterprise form

agjohnson:
I noticed a couple issues on casing, mainly "Enterprise" isn't a proper
noun and doesn't need to be capitalized.

----
📚 Documentation preview 📚: https://readthedocs-about--303.org.readthedocs.build/

### DIFF
--- a/content/pages/enterprise.html
+++ b/content/pages/enterprise.html
@@ -26,11 +26,11 @@
 
   <section>
     {% call basic.header_lead(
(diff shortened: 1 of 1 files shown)


ericholscher (approved):

ericholscher:
I think we use caps to discuss the plan types in other places on the pricing page, so I'd expect Enterprise plan would be the same?

PR #308, opened 2024-07-08
Post: enable addons by default

humitos:
Follows the plan we described in:
https://github.com/readthedocs/addons/issues/72#issuecomment-2200390705

----
📚 Documentation preview 📚: https://readthedocs-about--308.org.readthedocs.build/blog/2024/07/addons-by-default/


content/posts/addons-by-default.md
```diff
@@ -0,0 +1,100 @@
+title: Read the Docs Addons enabled by default
```
ericholscher, line 1: I feel like we need to lead with the build data deprecation, not Addons?

humitos: Hrm, I structured the post in this way because I'd assume that most users won't know what "build data deprecation" means and if they are using it or not.

I think the news "Addons enabled by default" and "What does it means?" and "How it affect my projects?" is a better way to structure it. That's why I went this way.

humitos: Also, I think that most of the users won't be directly affected by this (meaning, that it won't break their builds). I'm expecting a handful of users using the Sphinx context in a way that can hard break their docs and will require using the extension and/or contacting us to understand what happened.

humitos: In this context, the _news_ is that "Addons are being enabled by default" and it will have some implicates to the build process; not the other way around.

content/posts/addons-by-default.md
```diff
@@ -0,0 +1,100 @@
...
+
+- **July 29, 2024**: enable Read the Docs Addons for _all **new** projects_ created.
+- **October 7, 2024**: enable Read the Docs Addons for _**all projects**_.
+
+
+## How does it affect my projects?
+
+It's worth to note there are two big different types of changes that will affect your projects:
+at build time, and visually.
+Depending on the configuration of your project, it could be affected by one, both or none of them.
```
ericholscher, line 63:
```suggestion
There are two big different types of changes that will affect your projects:
at build time, and visually.
Depending on the configuration of your project, it could be affected by one, both or none of them.
```

content/posts/addons-by-default.md
```diff
@@ -0,0 +1,100 @@
...
+or the build could fail completely.
+
+In case you've found any difference in behavior after Read the Docs Addons has been enabled
+(see [How to opt-in to addons _now_](#how-to-opt-in-to-addons-now) to enable addons manually),
+you can try installing the [`sphinx-build-compatibility`](https://github.com/readthedocs/sphinx-build-compatibility/) extension we've created
+to allow keeping the "behind the scenes" manipulation behavior at build time while you can dig more on these changes and perform the migration gradually.
+
+### Visual changes
+
+Even if the build has completed successfully, there may be some visual changes you noticed.
```
ericholscher, line 79:
```suggestion
Even if the build has completed successfully, there may be some visual changes.
```

content/posts/addons-by-default.md
```diff
@@ -0,0 +1,100 @@
...
+Depending on your project's configuration, the documentation tool and the theme you are using, you will notice these differences.
+We plan to release a new version of our theme to keep the flyout menu at the bottom left integrated with the navbar soon,
+and we are talking to other theme authors to communicate these changes and work with them towards a better integration.
+Expect news around this visual changes in the following months.
+
+
+## How to opt-in to addons _now_
+
+Read the Docs Addons is already available for all users that want to start using it on their projects.
+In case they want to prepare themselves for the migration, or make usage of the new features from Read the Docs Addons, like the [DocDiff](https://docs.readthedocs.io/page/pull-requests.html) for example.
```
ericholscher, line 92:
```suggestion
In case they want to prepare themselves for the migration, or make usage of the new features from Read the Docs Addons, like the [DocDiff addon](https://docs.readthedocs.io/page/pull-requests.html).
```

ericholscher (commented):
I feel like this post needs a tl;dr about the context removal, and we need to be much more explicit around the fact it's happening. Addons being enabled is one thing, but the build changes are pretty big, and we want to highlight the extension, and explain more how it works, I think?

humitos:
I agree we can add a warning note, an extra paragraph with some bold text or similar alerting users about the build context changes. However, we don't want _all users to go and blindly install the extension_. I understand that a good tl;dr would be something like: "Read the Docs stopped adding extra context and manipulating configuration on Sphinx builds. If you are experimenting any issue after addons are enabled by default, read "How does it affect my projects?" section and try installing `sphinx-build-compatibility`"

On the other hand, I think going too deep technically on what exactly has changed, what are the exact variables that are not passed anymore, how to configure those variables by yourself, etc won't be valuable here and won't help most of the users and I think it will only confuse them more than help. They aren't already aware of those, and for those users needing them, the solution is just to install the Sphinx extension --we don't need to explain this on details to regular users.

humitos:
I went ahead and added a TL;DR small section at the beginning of the post. Besides, I added a small chunk of Python code that explains how to setup the canonical URL if they are using a custom domain, plus the `READTHEDOCS` variable in the context.

content/posts/addons-by-default.md
```diff
@@ -12,8 +12,19 @@ We are announcing the introduction of [Read the Docs Addons](https://about.readt
...
+> Read the Docs has stopped adding extra context and configuration on Sphinx builds.
+> This may affect your documentation builds.
+> If you are experimenting any issue with this change,
+> read [How does it affect my projects?](#how-does-it-affect-my-projects) section and
+> try installing [`sphinx-build-compatibility`](https://github.com/readthedocs/sphinx-build-compatibility) extension in your project to mitigate them._
+
+----
+
+Since Read the Docs started,
+every time it ran a build, a set of extra configuration were automatically added to make the integration easier.
```
ericholscher, line 27:
```suggestion
every time it ran a build, extra configuration options were automatically added to make integration easier.
```

ericholscher (approved):
Looks better. I'm fine not highlighting the context removal as much, but I do think it will break more than we think it will, especially with all the themes, etc.  👍

humitos:
Thanks for the review.

I'm happy to come back an update the post with some feedback received once we enabled it for new projects. We can also write a "Migration guide" in our documentation with more technical details if needed. We can also create an issue in our repository that links to this post and pin it; so users find it immediately when they go to open an new one about a related problem.

I targeted this post for next Monday, but we can merge it before that date if we are already happy with it. Let me know.

ericholscher:
Yea, I think making sure we have a few places where we're noting this is important. 👍 I'm fine shipping it on Monday.

PR #315, opened 2024-07-24
Update addons post to mention the theme release

humitos:
We released `sphinx-rtd-theme==2.1.0rc1` that keeps the old behavior regarding the flyout integration.

----
📚 Documentation preview 📚: https://readthedocs-about--315.org.readthedocs.build/


humitos:
I used ~~plan to realease~~ originally, but it seems that strike doesn't work on our blog for some reason.

ericholscher (approved):

PR #316, opened 2024-07-24
Pricing: hide "Basic" plan and show "Enterprise"

humitos:
I copied the text and features from http://localhost:8080/pricing/enterprise/

Screenshot_2024-07-24_16-21-26

Closes #289

----
📚 Documentation preview 📚: https://readthedocs-about--316.org.readthedocs.build/

PR #328, opened 2024-09-12
Pricing: mention specific features for Advanced

humitos:
Since we removed the Basic plan in #316, the phrase "Includes all Basic plan features" doesn't make sense anymore.

I removed that phrase and write down the features included instead.

----
📚 Documentation preview 📚: https://readthedocs-about--328.org.readthedocs.build/


ericholscher (approved):
Looks good. Might make the Advanced plan look a bit large, but seems fine for now.

PR #332, opened 2024-11-04
Point community to new dashboard

humitos:
We are already migrated on community, so I think we can start pointing people to the new dashboard already.

----
📚 Documentation preview 📚: https://readthedocs-about--332.org.readthedocs.build/

### DIFF
--- a/content/pages/privacy-policy.rst
+++ b/content/pages/privacy-policy.rst
@@ -105,7 +105,7 @@ Why do we collect this information?
 - We limit our use of your User Personal Information to the purposes listed in this Privacy Statement.
   If we need to use your User Personal Information for other purposes, we will ask your permission first.
   You can always see what information we have in your
-  `user account <https://readthedocs.org/accounts/edit/>`__.
+  `user account <https://app.readthedocs.org/accounts/edit/>`__.
 
 What information Read the Docs does not collect
 -----------------------------------------------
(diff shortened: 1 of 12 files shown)


content/pages/privacy-policy.rst
```diff
@@ -105,7 +105,7 @@ Why do we collect this information?
 - We limit our use of your User Personal Information to the purposes listed in this Privacy Statement.
   If we need to use your User Personal Information for other purposes, we will ask your permission first.
   You can always see what information we have in your
-  `user account <https://readthedocs.org/accounts/edit/>`__.
+  `user account <https://app.readthedocs.org/accounts/edit/>`__.
```
ericholscher, line 108: I think most of these probably need some similar solution to the dashboard links, where we have a way to link to a page that asks the user to go to .com or .org. It's probably out of scope for this PR though 🙃

ericholscher (commented):
This change looks good -- I think we're just blocked on https://github.com/readthedocs/readthedocs.org/issues/11209

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -106,12 +106,12 @@
...
 {% macro menu_log_in(with_signup=False) %}
   <div class="header">Log in</div>
-  <a class="item" data-analytics="community-login" href="https://readthedocs.org/dashboard/">
+  <a class="item" data-analytics="community-login" href="https://app.readthedocs.org/dashboard/">
     <i class="fad fa-people-group primary icon"></i>
     Read the Docs Community
 
     <p class="ui mini grey text">
-      <code>https://readthedocs.org</code>
+      <code>https://app.readthedocs.org</code>
```
ericholscher, line 114: I think we might not want to change this one, since it's more conceptually "the .org site" and not the specific URL.

agjohnson:
Already noted, but this is blocked on the notification to link users back to the legacy dasboard. With this, there would be no way to easily switch back.

readthedocs_theme/templates/includes/topnav.html
ericholscher: I wonder if we want to remove the `https://` here, to make it more concrete that it's conceptual?

ericholscher:
Seems safe to merge this now 👍

PR #337, opened 2024-11-21
Add basic landing page for any tool

ericholscher:
This is just a starting point with a generic config.

Fixes https://github.com/readthedocs/website/issues/301

----
📚 Documentation preview 📚: https://readthedocs-about--337.org.readthedocs.build/

### DIFF
--- a/content/includes/try-it-out-custom.html
+++ b/content/includes/try-it-out-custom.html
@@ -1,0 +1,117 @@
+<div class="ui container">
+    <div class="ui horizontal divider">Try it out</div>
+</div>
+
+<section id="try-it-out">
+    <div class="ui padded container">
+      <h2 class="ui huge center aligned header">
+        Upgrade your documentation with Read the Docs
+      </h2>
+
+      <div class="ui very padded centered grid">
+
+        <div class="ten wide computer sixteen wide tablet sixteen wide mobile column">
+          <div class="ui large left aligned basic vertical segment">
+
+            {# Stylized code blocks #}
+            {% call technical.code_block(title="<code>.readthedocs.yaml</code>", icon="fa-file") %}
+            {% markdown %}
+
+            ```yaml
+            # Read the Docs configuration file for Sphinx projects
+            # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+            # Required
+            version: 2
+
+            # Set the OS, Python version and other tools you might need
+            build:
+              os: ubuntu-24.04
+              tools:
+                python: "3.12"
+                nodejs: "20"
+                # You can also specify other tool versions:
+                # rust: "1.70"
+                # golang: "1.20"
+
+              commands:
+                  # Python
+                  # Install dependencies
+                  - pip install -r docs/requirements.txt
+                  # Build the site
+                  - make html
+                  # Copy generated files into Read the Docs directory
+                  - mkdir --parents $READTHEDOCS_OUTPUT/html/
+                  - cp --recursive docs/build/* $READTHEDOCS_OUTPUT/html/
+
+                  # Node
+                  # Install dependencies
+                  - cd docs/ && npm install
+                  # Build the site
+                  - cd docs/ && npm run build
+                  # Copy generated files into Read the Docs directory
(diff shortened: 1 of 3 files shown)


content/includes/try-it-out-custom.html
```diff
@@ -0,0 +1,117 @@
...
+        <div class="ten wide computer sixteen wide tablet sixteen wide mobile column">
+          <div class="ui large left aligned basic vertical segment">
+
+            {# Stylized code blocks #}
+            {% call technical.code_block(title="<code>.readthedocs.yaml</code>", icon="fa-file") %}
+            {% markdown %}
+
+            ```yaml
+            # Read the Docs configuration file for Sphinx projects
+            # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
```
humitos, line 22: We should follow the same pattern we already have in our "Add project" wizard form with a `tabmenu` and using the same examples: https://github.com/readthedocs/ext-theme/blob/0360c0e3a5a1219d97a3df035163482f5626d080/readthedocsext/theme/templates/projects/import_config.html#L26-L37

ericholscher: It seems if we're just putting the configs for all the tools here, it feels a bit weird? I think we just want to highlight we can run any tool, and not copy all the configs again here?

humitos: Ah, OK. Then, we should stick with one working example here only instead of using Python and Node in the same config outputting HTML files to the same directory.

humitos: I was thinking "more in general" where we are probably merging all the "Supported doctools" in only one page with all the examples together and a selector for the doctool. That way we don't need a specific page per doctool, and we can link to the page in our docs for each specific doctool -- that will explain things in deep.

readthedocs_theme/templates/includes/topnav.html
```diff
@@ -46,6 +46,7 @@
         <a class="item" href="{{ SITEURL }}/tools/sphinx">Sphinx</a>
         <a class="item" href="{{ SITEURL }}/tools/mkdocs">Mkdocs</a>
         <a class="item" href="{{ SITEURL }}/tools/jupyter-book">Jupyter Book</a>
+        <a class="item" href="{{ SITEURL }}/tools/custom">Any Tool</a>
```
agjohnson, line 49: Casing, and "any tool" feels in conflict with the other menu options

```suggestion
        <a class="item" href="{{ SITEURL }}/tools/custom">Other tools</a>
```

content/includes/try-it-out-custom.html
```diff
@@ -0,0 +1,117 @@
...
+            build:
+              os: ubuntu-24.04
+              tools:
+                python: "3.12"
+                nodejs: "20"
+                # You can also specify other tool versions:
+                # rust: "1.70"
+                # golang: "1.20"
+
+              commands:
```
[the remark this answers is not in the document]
ericholscher: That doesn't currently work for non-Sphinx versions, and I think the goal of this page is to promote custom commands, so I think build.commands is fine for now.

ericholscher:
Would love another review to get this out :+1:

humitos (approved):
I'm fine merging this PR for now. However, I _really_ want to talk more about our marketing pages and review what I said in https://github.com/readthedocs/website/pull/336#pullrequestreview-2450645615 and also in https://github.com/readthedocs/website/pull/307. The later is new content that I think we are currently missing.

I think we need a good re-structure and simplification of all these pages, so we don't keep adding more and more content because our current pages don't solve the problem. I really think we should have a lot fewer pages here with more concrete content that point users to our documentation pages for specific build options.

ericholscher:
@humitos Agreed -- I think we can use the docs for more of the iteration of various tools that we support. The docs have better SEO in general anyway, and the website can just have a listing. I'm imagining something like https://www.ethicalads.io/advertisers/#audiences but with a listing of supported tools linking to nice docs pages? 

But I do think it's important if we're gonna have a `Tools` section of the site to make sure we explicitly note supporting other tools. If we want to refactor the Tools section as a next step, I'm :100: on that.

PR #349, opened 2025-02-24
Add initial testomonials

ericholscher:
These are from public comments,
so I feel free safe linking them without permission.
I'll plan to ping folks and let them know we're adding them though.

Screenshot 2025-02-24 at 11 38 30 AM

Fixes https://github.com/readthedocs/website/issues/131


humitos (approved):
Looks good to me!

ericholscher:
🎉

PR #350, opened 2025-03-05
Add comparison index page.

ericholscher:
This cleans up the website footer.

FIxes https://github.com/readthedocs/website/issues/234

----
📚 Documentation preview 📚: https://readthedocs-about--350.org.readthedocs.build/

### DIFF
--- a/content/pages/comparisons/index.html
+++ b/content/pages/comparisons/index.html
@@ -1,0 +1,126 @@
+{# View this page at http://localhost:8080/comparisons/ #}
+
+<html>
+  <head>
+    <title>Read the Docs vs Other Documentation Platforms</title>
+    <meta name="description" content="Compare Read the Docs features with other documentation platforms like GitHub Pages, GitBook, and Cloudflare Pages to find the best solution for your documentation needs." />
+    <meta name="template" content="page" />
+    <meta name="slug" content="comparisons" />
+    <meta name="status" content="hidden" />
+  </head>
+  <body>
+
+  {% import "partials/basic.html" as basic %}
+
+  <section>
+    {% call basic.header_lead(
+    header="Compare Read the Docs with other platforms")
+    %}
+    <p>
+      While there are many options for hosting documentation,
+      Read the Docs is purpose-built for technical documentation with features that make both documentation authors and readers more productive.
+    </p>
+    <p>
+      Compare Read the Docs with other popular documentation platforms to find the best solution for your needs.
+    </p>
+    {% endcall %}
+  </section>
+
+  <section>
+    <div class="ui very padded container">
+      <div class="ui three stackable cards">
+
+        <div class="card">
+          <div class="content">
+            <a href="{{ SITE_URL }}/comparisons/github-pages/">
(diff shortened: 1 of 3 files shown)


content/pages/comparisons/index.html
```diff
@@ -0,0 +1,132 @@
...
+      Compare Read the Docs with other popular documentation platforms to find the best solution for your needs.
+    </p>
+    {% endcall %}
+  </section>
+
+  <section>
+    <div class="ui very padded container">
+      <div class="ui three stackable cards">
+
+        <div class="card">
```
humitos, line 33: Can we make each of these cards linkable instead? Does that make sense?

humitos (approved):
Makes sense to me.

content/pages/comparisons/index.html
ericholscher: I tried a couple approaches, and hit random annoying things with css. But I can try that.

content/pages/comparisons/index.html
```diff
@@ -0,0 +1,132 @@
...
+          <div class="column twelve wide computer sixteen wide tablet">
+            <div class="ui basic vertical huge segment">
+              <h2 class="ui center aligned header">
+                Ready to get started? <i class="fad fa-rocket icon"></i>
+              </h2>
+
+              <p>
+                Read the Docs is the all-in-one solution that provides everything you need
+                for hosting great documentation, from analytics to versioning.
+                Try it today and see the difference!
```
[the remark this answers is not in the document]
ericholscher: I don't think anyone will meaningfully read this text, we just needed something short & sweet. It's basically just for search engines.

ericholscher:
Alright, I think I got something nicer:

Screenshot 2025-03-28 at 12 23 45 PM

I had to use explicit CSS to make the height of the images the same -- is there a way to do this with semantic UI? I wasn't sure how to target it with an external CSS file for something small like this.

humitos:
I like it. The whole card is clickable now and it moves a little when you hover, which is a good visual differentiator 👍🏼 

The inline CSS is a problem, tho, I'm not sure why you weren't able to use a class and put it outside the HTML. I'm sure I had a similar problem in other places already and I fixed it _somehow_. Maybe you can check the other files.

agjohnson:
Yup, all images should be edited to be the same proportions going in.

content/partials/basic.html
```diff
@@ -107,3 +107,38 @@ <h3 class="ui {{ size }} header">
...
+  :param title: Card title/platform name
+  :param meta: Short description below title
+  :param description: Main card description
+#}
+{% macro comparison_card(url, logo_url, logo_alt, title, meta, description) %}
+  <a class="card" href="{{ url }}">
+    <div class="content">
+      <div class="ui basic padded segment" style="height: 100px; display: flex; align-items: center; justify-content: center;">
+        <div class="ui small centered image">
+          <img src="{{ logo_url }}" alt="{{ logo_alt }}" style="max-height: 50px; width: auto;">
```
agjohnson, line 128: You should instead edit the images to get similar proportions and alignment.

`.ui.small.image` already controls the size and centering of the image, inline styles or a statically sized segment shouldn't be used to dictate this.

PR #351, opened 2025-03-05
Simplify nav and remove links to other workflow pages for now.

ericholscher:
- **Simplify nav and remove links to other workflow pages for now.**
- **Migrate to bottom nav**

----
📚 Documentation preview 📚: https://readthedocs-about--351.org.readthedocs.build/

### DIFF
--- a/content/pages/features.html
+++ b/content/pages/features.html
@@ -188,12 +188,10 @@
         </div>
       {% endblock authoring_details %}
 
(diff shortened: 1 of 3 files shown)


humitos (approved):
I like this approach more. We have only one page to point users and from there we point them to more specific pages in case they want to understand more about a particular feature/workflow/etc 👍🏼

PR #352, opened 2025-03-25
fix(jekyll): fix bundler package name

ReenigneArcher:
The correct package is `bundler` not `bundle`.

I also made a PR here: https://github.com/readthedocs/ext-theme/pull/574


ericholscher (approved):
Thanks for the fix!

PR #377, opened 2026-03-09
Newsletter for March

humitos:
This post was all generated by AI. I will review it after pushing it and make sure it's correct. This is just a test to try to revive the newsletter without a lot of effort.

* Related https://github.com/readthedocs/meta/issues/111


ericholscher:
I think if we want to move forward with this, we should use AI to pull together a listing of things to cover, but not actually write the content. I think we owe our users to write the actual content, but I do think the content gathering is a great AI task. 

We moved to a quarterly newsletter on ads, and that feels about right. Each issue has real major changes, which is nice.

ericholscher:
Would be good to get a Q2 newsletter out this month 🤔

PR #390, opened 2026-04-14
Update internal links to use URL paths instead of file references

ericholscher:
These 2 files were showing in every files changed view because the links general the full URL including domain, which varies in every PR.


humitos (approved):
Looks good.

Isn't there a way to make an internal link similar to `:doc:` in Sphinx? Otherwise, if we move these pages these links will break.

ericholscher:
Yea, I think if we move them, we just catch them with 404 checking.

humitos:
> I think if we move them, we just catch them with 404 checking.

Do we have a 404 checking here? 🤔

PR #394, opened 2026-04-21
Post: deprecation of `python.install.method: setuptools`

humitos:


content/posts/deprecate-setuptools-install.md
```diff
@@ -0,0 +1,74 @@
...
+```yaml
+version: 2
+
+python:
+  install:
+    - method: setuptools
+      path: .
+```
+
+you should update it to use `pip` instead:
```
agjohnson:
```suggestion
These projects should update their configuration file to use `pip` instead:
```

content/posts/deprecate-setuptools-install.md
```diff
@@ -0,0 +1,74 @@
...
+image_credit: Photo by <a href="https://unsplash.com/@hnhmarketing?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Hunter Haley</a> on <a href="https://unsplash.com/photos/four-handheld-tools-on-board-s8OO2-t-HmQ?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Unsplash</a>
+      
+
+We are announcing the deprecation of `python.install.method: setuptools` in the `.readthedocs.yaml` configuration file.
+
+This option uses `python setup.py install`, which is deprecated in the Python packaging ecosystem.
+Read the Docs will remove support for it on **Wednesday, July 1st, 2026**.
+After this date, projects still using `python.install.method: setuptools` will fail their builds until they are updated.
+
+If your project is installing a local Python package with this configuration:
```
agjohnson:
```suggestion
Projects using this install method will have a configuration file that looks like this:
```

agjohnson (approved):
We probably also want this information in our docs and should link there for more information. This blog post is generally pretty in depth, the content feels more like it should be in our docs. I don't think it's bad for it to live in multiple places though, as long as the sources match.

content/posts/deprecate-setuptools-install.md
```diff
@@ -0,0 +1,74 @@
+title: Deprecation of python.install.method: setuptools
+date: 2026-04-21
```
humitos, line 2: 📓 Update the date before publishing.

humitos:
> This blog post is generally pretty in depth, the content feels more like it should be in our docs.

We usually write the technical details in these posts about how to migrate when we deprecate something. The idea is to point everybody here as the source of truth (from the email, the docs, social media, etc). The `setuptools` content from other places (like docs) is going to be removed. It's already marked as deprecated.

PR #395, opened 2026-04-23
Trim try-it-out YAML example to essentials

ericholscher:
Clean up the code sample to be more glanceable, and not as verbose. Marketing pages should get to the value. 

---

_Generated by [Claude Code](https://claude.ai/code). Session: https://claude.ai/code/session_01RauZs2WN5Gunnp3LQ5jhBD_


agjohnson (approved):
I'm glad to have shorter examples, I had similar feedback initially. I think examples on the site are more for illustration only, they should be as short as possible to communicate ease of use. Documentation examples are where we should expect users to copy and paste from.

PR #404, opened 2026-04-29
docs: fix logo shift in sticky comparison/pricing table header

ericholscher:
## Summary

When scrolling past the comparison/pricing feature table, the Mintlify logo (and any other logo sized as a percentage of its `<th>`) visibly shifted inward. Root cause: Fomantic's JS sticky module sets `position: fixed` on the `<thead>` once it scrolls into view, which detaches the thead from the table layout and collapses the columns to their declared "wide" widths (`eight wide` + `two wide` + `two wide` = 50% + 12.5% + 12.5%). The Mintlify `` scales with the cell, so it shrinks and shifts.

Switch the sticky behavior to native CSS `position: sticky` on the `<th>` cells (not the `<thead>`). Sticky cells stay in flow, table layout is preserved, and column widths don't recalculate — so the logos no longer move.

Affects both `content/partials/comparisons.html` and `content/partials/pricing.html`, which use the same `feature_table` macro.

> Note: this branch also includes the `build: unbreak pre-commit in dev environments` commit, which is the same commit on top of #403. When either PR merges, the other will rebase cleanly.

## Test plan

- [x] Load `/comparisons/mintlify/` (and any other comparison page) and scroll past the table header — the Mintlify logo should stay centered in its column with no shift
- [x] Same on the pricing page (`/pricing/`) — the plan buttons in the sticky header should stay in their columns
- [ ] Verify on narrow viewports that table layout still works (sticky cells need a background to avoid bleed-through; set to `#fff`)
- [ ] Visually confirm no z-index regressions against any modals/menus on these pages

Generated by AI.

---
_Generated by [Claude Code](https://claude.ai/code/session_012AJaFmnrTPtNbjfNsyk8Cg)_


ericholscher:
This actually fixes the existing glitch where it stutters when you scroll into the box as well!

Great job AI <3

ericholscher:
Tested this on Chrome and it's smooth like butter.

src/sui/themes/rtd-site/globals/site.overrides
```diff
@@ -1,3 +1,13 @@
+// Sticky on the <th> cells (not <thead>) so the table layout is preserved
+// while pinned -- a position:fixed/sticky <thead> detaches from the table and
+// collapses columns to their declared "wide" widths, shifting the logos.
+#feature-table thead.sticky-header th {
```
agjohnson, line 4: This should stick to the patterns and conventions used in SUI already instead of mixing in competing patterns. Using ids in selectors makes for one off rules that aren't reusable.

Making this a rule like .ui.sticky.header.table thead th {} instead will follow the patterns used in SUI and is a reusable, general solution.

src/sui/themes/rtd-site/globals/site.overrides
```diff
@@ -1,3 +1,13 @@
...
+  position: sticky;
+  top: 0;
+  z-index: 1;
+  background-color: #fff;
```
agjohnson, line 8: Is this color intentional? I prefer the base styles for thead here, it's distinct from a normal row. If this is important, it doesn't match the tfoot now and looks unfinished.

Image

agjohnson (commented):
I'm guessing the sticky FUI module needs something extra here, I'd be surprised if this can't easily be solved withing FUI though. I doubt the llm would give a workable fix though, it's gravitating towards breaking us out of FUI already.

Hand rolling sticky positioning is going to be hard, there is janky behavior waiting here -- tablet viewport widths clip the sticky header for example:

Image

humitos:
I think there is something else here to investigate. Why it does happens with Mintlify logo but not with the other ones. We have more comparison pages that are working fine. Is it because of the size of the SVG? Maybe the solution is just to re-scale that file instead of making complex CSS changes 😄

ericholscher:
The current behavior is definitely broken -- it always skips the first row item on scroll.

agjohnson:
In general, we should indeed edit graphics files to the same resolution and padding as we use them, instead of using all the images raw. It's part of the cause here but mostly just good practice.

But like I said, I think there is just something else going on here and spending time looking into FUI is a good use of time. There's a note on the sticky module that sticky on tables requires a fixed table layout.

https://fomantic-ui.com/modules/sticky.html#initializing

Sticky has to break the `thead` out of the table, which is what resize all the other cells and is likely the problem to both issues.

FUI also already supports sticky headers in tables, from 2.9.0:

https://fomantic-ui.com/collections/table.html#stuck

I put together a quick change here:

https://github.com/readthedocs/website/pull/408/changes/4b1d13132dc915f9b40868e08e180eff555cb92e

Building isn't working atm, NPM is not cooperating with me. This change requires at least an upgrade on FUI, but pretty much everything here needs upgraded to match ext-theme too -- webpack and friends, the webpack config, and a less ancient Nodejs.

- https://github.com/readthedocs/website/pull/408

PR #408, opened 2026-05-07
Upgrade FUI and use stuck head table instead of sticky module on tables

agjohnson:
Big upgrade of everything so that we can use FUI 2.9+. This repo has lagged way behind ext-theme, this is largely matching the dependencies there.

The underlying cause for this PR is fixing the sticky row jumping out of table element and fixes issue with images in a sticky module row.

With FUI 2.9 and stuck tables, the feature tables act how we want. The webpack upgrade was necessary to build the assets with the new libraries.


agjohnson:
I started from testing FUI 2.9+ without upgrading everything, but very quickly building and icons broke due to changes in 2.9.4.

There is a lot here upgrading tools as this was a big upgrade we've avoided, But it does now match ext-theme configuration pretty closely. The fix for the display issue is small in comparison, just

[4b1d131](https://github.com/readthedocs/website/pull/408/commits/4b1d13132dc915f9b40868e08e180eff555cb92e)

humitos:
It's looking good.

The build fails due to small validation errors:
```
error: "target" attribute cannot be used in this context: requires "href" attribute to be present (attribute-misuse) at output/drafts/pages/examples/marketing.html:675:21:
```

webpack.config.mjs
agjohnson, line 1: Git is confused here, this diff is bad. For a better view on what changed:

https://github.com/readthedocs/website/compare/agj/stuck-tables-bad-diff?expand=1#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166f

agjohnson:
Yeah, I ran out of steam here. The fix was small though.

All of these changes come from ext-theme, so are pretty vetted. There might be some minor issues kicking around though, this upgrade has been needed for a long time.

davidfischer:
I pushed a small README update. The fonts do not quite look right and I see font download errors in the browser console (FF 150.0.2 on Linux). It's falling back.

davidfischer:
I pushed a webpack fix for the Lato font.

agjohnson:
I tried to make the publicPath `./` instead of `../css` and it does not work. There is probably something here but I'll say it's safe to punt on this.

davidfischer:
:shipit:

PR #409, opened 2026-05-12
Offite changes

ericholscher:


davidfischer (approved):
There's some weird spacing changes and linebreaks in the template that don't affect the HTML display. Verbiage-wise, these changes are an improvement.

PR #413, opened 2026-05-14
Remove all monospace headers

davidfischer:
Headers should just use the `header` class.

This is separate from the fact that there's a font issue with monospace (the font stack isn't quite right).


agjohnson (approved):
`Looks good to me`

PR #416, opened 2026-05-14
Updated pre-commit configuration

davidfischer:
Remove broken symlinks and don't rely on common pre-commit.

The only thing I don't like here myself (and since you'll run it way more than me I'll let you decide) is that the `check-build-assets` check is slow and for a full check can take the better part of a minute. It does a full rebuild of the JS/CSS and does essentially the same thing as ext-theme.

## Reviewing
For reviewing this PR, you might want to review the first commit separately than the 2nd commit. The first one just updates the pre-commit config. The second one (a20e5ac) runs `uv run pre-commit run --all-files` across the whole repo. This makes it so everything is linted/pre-committed.


ericholscher (approved):
Looks great!

agjohnson (approved):

PR #424, opened 2026-05-14
More anthony pricing header changes

davidfischer:


davidfischer:
I've reviewed Anthony's changes and they look good on desktop and mobile.

PR #425, opened 2026-05-14
Upgrade the tool versions in config examples

davidfischer:


content/includes/try-it-out-docusaurus.html
```diff
@@ -20,9 +20,9 @@ <h2 class="ui huge center aligned header">
             ```yaml
             version: 2
             build:
-              os: "ubuntu-22.04"
```
humitos, line 23: We should probably use 26 at this point

humitos (approved):
Looks good 🚢

ericholscher:
Merging here because it's better than what we have, but we can do a 26.06 upgrade after if we want.

PR #428, opened 2026-05-30
Using NPM minimum age for dependencies

agjohnson:
This enforces package age before installation for local package
installation. It does not address dependabot yet though, there is a
separate fix needed for this.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 engine-strict=true
+min-release-age=14400
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,2 @@
 engine-strict=true
-min-release-age=14400
+min-release-age=14
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,3 @@
 engine-strict=true
-# Don't install packages younger than 14d, also set in .ncurc.mjs
+# Don't install packages younger than 14d
 min-release-age=14
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 engine-strict=true
+# Don't install packages younger than 14d, also set in .ncurc.mjs
 min-release-age=14
--- a/.tool-versions-example
+++ b/.tool-versions-example
@@ -1,3 +1,3 @@
-nodejs 22.11.0
+nodejs 24.16.0
 python 3.10.8
 uv 0.7.8
--- a/README.rst
+++ b/README.rst
@@ -29,7 +29,7 @@ built using a number of Node dependencies.
 
 You will need the following requirements:
 
-* Node.js version ``22``
+* Node.js version ``24``
 * NPM version ``11.10``
 * Python version ``3.10`` (any release ``>=3.6`` works)
 * uv (Use ``asdf`` or see https://github.com/astral-sh/uv for installation instructions)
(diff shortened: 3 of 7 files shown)


.npmrc
```diff
@@ -1 +1,2 @@
 engine-strict=true
+min-release-age=14400
```
humitos: We could use a comment here about how much time this is.

humitos (commented):
Should we do the same for addons?

.npmrc
agjohnson: It's in days anyways, the unit was a bit ambiguous. I added a note either way though.

agjohnson:
Yeah, I didn't get to addons last week. Just put up a PR there

agjohnson:
I've simplified this a bit and just pinned Node 24 as the required. Node 22 was a bit old and I hit issues on addons trying to get dependencies to work with Node 22. Node 24.16 includes the npm version we want as well, so a good default version generally.

README.rst
```diff
@@ -115,17 +116,11 @@ Linting is required for every pull request, skipping this step can cause the
 build to fail if your formatting doesn't match the intended output from
 ``prettier``.
 
-To run linting checks, which will only report errors:
+All linting is set up to run via ``pre-commit``:
```
humitos, line 119: Is this true? I've never used pre-commit in this repository 😅

humitos: Also, I don't think that pre-commit does the `npm run format` either.

humitos (commented):
This looks good, but I think we should keep the lint/format commands to be run via `npm`. This repository doesn't have pre-commit configured as far as I can tell.

.npmrc
```diff
@@ -1 +1,3 @@
 engine-strict=true
+# Don't install packages younger than 14d, also set in .ncurc.mjs
```
humitos: It seems we need to bring this file in, right? https://github.com/readthedocs/addons/pull/702/changes#diff-d35686a0117b5567e16d7f08d3ee0027fa9cc02b4ef4d3ec23305724228ced21

humitos:
> This looks good, but I think we should keep the lint/format commands to be run via `npm`. This repository doesn't have pre-commit configured as far as I can tell.

Sorry, I've got confused with addons repository --too many similar PRs 🙃

humitos (approved):

.tool-versions-example
ericholscher, line 1: Should we remove the example here, and just ship the real file?

agjohnson: I've been +1 on this for a while, but I think there is some reason others didn't want this defined in the repository.

.npmrc
agjohnson: I actually noticed a modern ncu just pick up the setting here anyways. This isn't using ncu yet, I might explore that for website/ext-theme next. But seems like with a modern ncu we can also skip configuring it. I'll just drop the comment here.

```suggestion
# Don't install packages younger than 14d
```

.tool-versions-example
agjohnson: I'll save this for later, it might be good to do across repositories. It's unrelated to this change though, so I don't want to hold this up.

- https://github.com/readthedocs/meta/issues/226

PR #429, opened 2026-05-30
Add notice about june maintenance window for database upgrade

agjohnson:
Not sure if this is even needed, it looked like downtime was just going
to be a blip if anything. If we want to go for the major version upgrade
we should go ahead with this and maybe reframe the potential for
downtime.


stsewd (approved):

agjohnson:
It ended up being a blip on community, so we don't really need this

PR #431, opened 2026-06-27
Add platform chooser popup web component

agjohnson:
A reusable element to intercept a link out to our application and provide a
basic selection between both platforms for the same link without having to
hardcode this into the content we're writing. This replaces the confusing
pattern of:

```
To update your account email address, visit your account settings:

- On Read the Docs Community (https://readthedocs.org), <a href="..">go here</a>
- On Read the Docs Business (https://readthedocs.com), <a href="..">go here</a>
```

With:

```
To update you account email address,
<readthedocs-platform-chooser next="..">
<a>visit your account settings dashboard</a>
</readthedocs-platform-chooser>
```

image


humitos (approved):
Great idea!

I assume this will always render like a button, right? I mean, it's not an inline element that we can use just as a regular link inside a sentence.

agjohnson:
It does also work as an inline element, this is just showing what it looks like to use a button. I have a blog post example that will use this as both.

I'll fix the checks but if those look good I'll merge and get the blog post up next.

agjohnson:
From

- #432 

https://readthedocs-about--432.org.readthedocs.build/blog/2026/07/dashboard-dark-theme/

